PR #63, opened 2026-03-03
DC-137: Add document verification flow with Socure DocV adapter

necampanini:
## Summary

Adds a document verification (DocV) flow using Socure's SDK as an alternative identity-proofing path. When the backend returns `documentVerificationRequired` after ID proofing submission, the user is guided through document capture → verification polling → result routing.

**What's new:**
- Vendor-agnostic DocV adapter interface with Socure and mock implementations — app code never touches the SDK global directly
- Three-phase page state machine: interstitial → capture → verification pending, with sessionStorage persistence for mobile tab recovery (Socure opens a new tab on mobile)
- Off-boarding page for verification failures with contact/retry options
- API hooks: `useStartChallenge` (mutation) and `useVerificationStatus` (polling with exponential backoff)
- Zod schemas for the new endpoints

**Also in this PR:**
- Updated DC and CO locale CSVs from Google Sheets — introduces `stepUpDisclaimer`/`stepUpFailure` namespaces and updated content for several existing keys
- Fixed `generate-locales.js` to handle hyphenated CSV page names (e.g. "Step-up Disclaimer") — both mapped the new names in config and added identifier sanitization as defense-in-depth
- Defensive guard in `IdProofingForm` if backend sends `documentVerificationRequired` without a `challengeId`
- Updated 4 tests whose expected strings changed with the new locale content

## What's deferred
- **~27 hardcoded i18n strings** in doc-verify/off-boarding components all have `// TODO` comments — content team hasn't authored these keys in the Google Sheet yet
- **CSP headers** (`frame-src`, `child-src`, `Permissions-Policy: camera`) need updating before real SDK integration — currently running mock adapter only
- **SDK version pinning / SRI** for the Socure bundle.js — requires sandbox access
- **Unhandled launch rejection** in `DocVerifyCapture` — only affects real SDK path

## Test plan
- [x] Locale generation produces valid TS with new `stepUpDisclaimer`/`stepUpFailure` namespaces
- [x] 351 frontend tests pass (including 1 new guard test for missing `challengeId`)
- [x] 424 backend tests pass
- [x] ESLint: 0 errors
- [x] Cross-model review by Gemini and Codex — findings addressed or documented as deferred

### DIFF
--- a/src/SEBT.Portal.Web/src/features/auth/components/doc-verify/DocVerifyPage.tsx
+++ b/src/SEBT.Portal.Web/src/features/auth/components/doc-verify/DocVerifyPage.tsx
@@ -5,18 +5,17 @@ import { useCallback, useEffect, useRef, useState } from 'react'
 
 import { Alert } from '@/components/ui'
 
+import {
+  SK_ALLOW_ID_RETRY,
+  SK_CHALLENGE_ID,
+  SK_SUB_STATE,
+  SubState
+} from '@/features/auth/components/doc-verify/sessionKeys'
 import { useStartChallenge } from '../../api'
 import { createDocVAdapter, type DocVAdapter, type DocVAdapterConfig } from './adapters'
 import { DocVerifyCapture } from './DocVerifyCapture'
 import { DocVerifyInterstitial } from './DocVerifyInterstitial'
 import { VerificationPending } from './VerificationPending'
-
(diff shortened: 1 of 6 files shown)


noxmwalsh:
This is probably more of a general question how flexible we want the app to be, but do we want frontend routes that aren't relevant to a state (CO wouldn't be using Socure, for example) to be disabled or redirect elsewhere?

PR #73, opened 2026-03-10
DC-167(Feat): Exclude sibling repos from format checks and hide sensitive fields via Feature Flag

adbergen:
This pull request introduces new feature flags for controlling the visibility of sensitive information (case numbers and card last 4 digits) in the application, and integrates these flags into both the frontend and backend. It also updates related tests to support the new flags and ensures that formatting checks in the pre-commit hook exclude external projects. The most important changes are summarized below:

**Feature flag integration (frontend and backend):**
* Added new feature flags (`show_application_number`, `show_case_number`, `show_card_last4`) to `appsettings.json`, `appsettings.dc.json`, and the test mock flags, allowing these features to be toggled per environment. [[1]](diffhunk://#diff-b292786a8ea5c76439808bfba75f8de952b57016b474756ef5e672c80c3e16c6R88-R90) [[2]](diffhunk://#diff-24dfee0f218b88c76bd7ab7e223bdf519c10279173015cd2a4dbdb3784fda9ebR1-R7) [[3]](diffhunk://#diff-3158ec701654dd5f6ae4c91ca18989a5a1e239bde8956342262f9bc40e958ff6L39-R42)
* Updated the `ApplicationsSection` and `ChildCard` components to conditionally display case numbers and card last 4 digits based on the corresponding feature flags, using the `useFeatureFlag` hook. [[1]](diffhunk://#diff-79464c440b61cac430d3e3943cc3bf6698f841c4e10546502bca6b58746c89f1R5-R6) [[2]](diffhunk://#diff-79464c440b61cac430d3e3943cc3bf6698f841c4e10546502bca6b58746c89f1R35) [[3]](diffhunk://#diff-79464c440b61cac430d3e3943cc3bf6698f841c4e10546502bca6b58746c89f1L42-R45) [[4]](diffhunk://#diff-bdf7ada92262457200bebf38d8de7672eb34f91b3f66ad2375600127d760b803R6-R7) [[5]](diffhunk://#diff-bdf7ada92262457200bebf38d8de7672eb34f91b3f66ad2375600127d760b803R31) [[6]](diffhunk://#diff-bdf7ada92262457200bebf38d8de7672eb34f91b3f66ad2375600127d760b803L92-R95)

**Test improvements:**
* Refactored tests for `ApplicationsSection` and `ChildCard` to use a `FeatureFlagsContext` provider, enabling tests to simulate different flag configurations. Added new tests to verify that case numbers and card numbers are hidden when their flags are off. [[1]](diffhunk://#diff-67df28e17835574efc18cfae662b126c8fe0aca51c42f6c7be0ed48fb8af0dd6R4-R6) [[2]](diffhunk://#diff-67df28e17835574efc18cfae662b126c8fe0aca51c42f6c7be0ed48fb8af0dd6R43-R81) [[3]](diffhunk://#diff-67df28e17835574efc18cfae662b126c8fe0aca51c42f6c7be0ed48fb8af0dd6L124-R157) [[4]](diffhunk://#diff-547faebb2bf6e7d54debb8c788ba91b63b281e30f20352bdc192d788f5771944R5-R7) [[5]](diffhunk://#diff-547faebb2bf6e7d54debb8c788ba91b63b281e30f20352bdc192d788f5771944R34-R62) [[6]](diffhunk://#diff-547faebb2bf6e7d54debb8c788ba91b63b281e30f20352bdc192d788f5771944R205-R221)

**Developer tooling:**
* Updated the `.husky/pre-commit` script to exclude external sibling repositories from `.NET` formatting and analyzer checks, preventing errors when these projects are not present.


src/SEBT.Portal.Api/appsettings.dc.json
```diff
@@ -0,0 +1,7 @@
+{
+  "FeatureManagement": {
```
noxmwalsh: So this was a thought I was wondering: do we want these state-specific FF settings being persisted in VC?      The assumption is that in the future other states will be working with this repo, so it might open up confusion and the possibility that another states settings could be accidentally busted if modified. 

I think it might be ok if it stays as a .example file instead, but I'm curious what you think

src/SEBT.Portal.Api/appsettings.json
```diff
@@ -85,6 +85,9 @@
   },
   "FeatureManagement": {
     "email_dob_opt_in": false,
+    "show_application_number": true,
+    "show_case_number": true,
+    "show_card_last4": true,
```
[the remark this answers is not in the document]
adbergen: The defaults are intentionally true. These are base settings for all states.

src/SEBT.Portal.Api/appsettings.dc.json
adbergen: Good point! I think keeping it in VC is the right move here. These are feature flags, not secrets, so there's no risk having them checked in. And having them in git means we get PR review on any flag changes and full history of what was on/off when. If we go .example we lose that visibility and add a manual setup step for every dev. Though I do think maybe this should be a team discussion/decision, since client related.

This ties into [DC-142](https://codeforamerica.atlassian.net/browse/DC-142) where we're using these flags to control what DC users see. Keeping the overrides in VC means the flag state is tied to the code that consumes them.

noxmwalsh (approved):

PR #78, opened 2026-03-11
DC-138: Add Socure DocV backend integration

necampanini:
Adds the backend for Socure DocV document verification, following the frontend flow built in DC-137. When Socure's risk assessment determines a user needs additional identity proofing, this branch creates a document verification challenge, generates a Socure DocV session, receives the asynchronous verification result via webhook, and updates the user's IAL level on success.

Contains new domain model, four API endpoints, Socure HTTP client, webhook handler, EF Core migration, and full test coverage.

### Domain model

- `DocVerificationChallenge` entity with a strict state machine: `Created -> Pending -> Verified | Rejected | Expired`
- Terminal states are immutable — a verified user can't be downgraded by a late webhook
- Challenges use opaque GUIDs as public IDs (IDOR prevention); all reads scoped by `(publicId, userId)`
- Check-on-read expiration: if a pending challenge is past its `ExpiresAt`, the status query transitions it to Expired on the fly

### API surface (4 endpoints)

| Method | Path | Auth | Purpose |
|--------|------|------|---------|
| POST | `/api/id-proofing` | JWT | Submit ID data for Socure risk assessment. Returns `matched`, `failed`, or `documentVerificationRequired` with a `challengeId` |
| GET | `/api/challenges/{id}/start` | JWT | Returns DocV transaction token + URL for the frontend SDK |
| GET | `/api/id-proofing/status?challengeId={id}` | JWT | Poll for verification result (`pending`, `verified`, `rejected`) |
| POST | `/api/socure/webhook` | Anonymous (signature-validated) | Receives Socure `evaluation_completed` events |

### Socure integration

- `ISocureClient` interface with two implementations:
  - `StubSocureClient` — deterministic responses for local dev (default)
  - `HttpSocureClient` — real Socure Orion API calls, tested against sandbox
- Single-call design: the initial risk assessment call also generates the DocV session token, which is stored on the challenge for later retrieval (no second Socure call needed at challenge start)
- Configuration externalized via `SocureSettings` — API key, base URL, webhook secret, workflow name, enrichment provider name, challenge expiration

### Webhook handling

- Bearer token authentication via constant-time comparison (`CryptographicOperations.FixedTimeEquals`)
- Idempotency: stores Socure `event_id` on the challenge; duplicate events are silently acknowledged
- Correlation: primary lookup by `SocureReferenceId`, fallback to `EvalId`
- Always returns 200 to Socure to prevent retries; failures are logged
- Startup guard: non-Development environments fail to start if webhook secret is missing

### Frontend changes (minimal)

- `DocVerifyPage` now reads `challengeId` from URL query param instead of `sessionStorage` (PII handling compliance)
- `IdProofingForm` passes structured `DateOfBirth` DTO
- `allowIdRetry` derived from status API response (not URL-tamperable)
- Added `useVerificationStatus` hook test

### Infrastructure

- EF Core migration: `AddDocVerificationChallengesTable` with indexes on `PublicId`, `SocureReferenceId`, `EvalId`, `UserId+Status`
- `DependencyFailedResult` mapped to 502 Bad Gateway in `MvcResultExtensions`
- Named `HttpClient` registration for Socure with configurable base URL and auth headers

## Open questions

1. **Socure as shared service vs. plugin** — Currently built as shared infrastructure. If Socure ends up being DC-only, could this move into the state connector? The `ISocureClient` interface makes this a contained refactor.
2. **Challenge expiration duration** — Defaulted to 30 minutes (configurable via `Socure:ChallengeExpirationMinutes`). Is 30 minutes appropriate, or should we consult with the state partner?
3. **Production workflow name** — The sandbox uses `consumer_onboarding`, which appears to be a dev onboarding workflow. The production workflow may differ. This is config-driven (`Socure:Workflow`), so no code change needed — just flagging for deployment planning.
4. **Token refresh after verification** — When the webhook marks a user as verified, the frontend needs to refresh its JWT to pick up the updated `IdProofingStatus` claim. The existing `RefreshTokenCommand` handles this, and the frontend polls status then navigates — but the explicit "refresh token after verified" step isn't wired in the frontend yet. 

### Running the test suite

```bash
pnpm api:test
```

Key test files:
- `SubmitIdProofingCommandHandlerTests` — 6 scenarios including validation, active challenge reuse, Socure outcome routing
- `StartChallengeCommandHandlerTests` — 6 scenarios including IDOR prevention, state validation, idempotent starts
- `ProcessWebhookCommandHandlerTests` — 6 scenarios including signature rejection, idempotency, terminal state protection
- `GetVerificationStatusQueryHandlerTests` — 4 scenarios including check-on-read expiration
- `HttpSocureClientTests` — unit tests for request building, response parsing, enrichment extraction

### Happy path

1. `POST /api/id-proofing` -> `SubmitIdProofingCommandHandler` -> `ISocureClient.RunIdProofingAssessmentAsync()` -> returns `documentVerificationRequired` + `challengeId`
2. `GET /api/challenges/{id}/start` -> `StartChallengeCommandHandler` -> returns stored DocV token (no second Socure call)
3. Socure sends webhook -> `WebhookController` -> `ProcessWebhookCommandHandler` -> transitions challenge to Verified, updates user IAL
4. `GET /api/id-proofing/status?challengeId={id}` -> `GetVerificationStatusQueryHandler` -> returns `verified`

### Smoke tests (requires Socure sandbox API key)

```bash
dotnet test --filter "FullyQualifiedName~HttpSocureClientSmokeTests"
```

These are `[Fact(Skip = "...")]` by default. To run against the real sandbox, remove the skip and set `SOCURE_API_KEY` env var.

## Local setup for sandbox testing

By default, `Socure:UseStub` is `true` — the app uses `StubSocureClient` and no credentials are needed. To test against the real Socure sandbox:

### 1. Get credentials from the Socure RiskOS dashboard

- **API key:** 
- **SDK key:** 
- **Webhook secret:** This is a shared bearer token — you configure it on both sides. Set a secret string in your app config, then configure the same string as the `Authorization: Bearer` header in Socure's webhook settings. The exact Socure dashboard flow for configuring outbound webhook auth hasn't been confirmed yet.

### 2. Set user secrets for the API project

```bash
cd src/SEBT.Portal.Api
dotnet user-secrets set "Socure:UseStub" "false"
dotnet user-secrets set "Socure:ApiKey" "<your-sandbox-api-key>"
dotnet user-secrets set "Socure:WebhookSecret" "<your-shared-bearer-token>"
```

### 3. Set the frontend SDK key

In `src/SEBT.Portal.Web/.env.local`:
```
NEXT_PUBLIC_SOCURE_SDK_KEY=<your-sandbox-sdk-key>
```

### DIFF
--- a/src/SEBT.Portal.UseCases/IdProofing/ProcessWebhook/ProcessWebhookCommandHandler.cs
+++ b/src/SEBT.Portal.UseCases/IdProofing/ProcessWebhook/ProcessWebhookCommandHandler.cs
@@ -26,6 +26,9 @@ public class ProcessWebhookCommandHandler(
     ILogger<ProcessWebhookCommandHandler> logger)
     : ICommandHandler<ProcessWebhookCommand>
 {
+    private static string SanitizeForLogging(string? value) =>
+        value?.Replace("\r", string.Empty).Replace("\n", string.Empty) ?? string.Empty;
+
     public async Task<Result> Handle(
         ProcessWebhookCommand command,
         CancellationToken cancellationToken = default)
@@ -54,7 +57,7 @@ public class ProcessWebhookCommandHandler(
             // contract with Socure has changed and all webhooks are being silently dropped.
             logger.LogError(
                 "Webhook received but no challenge found for ReferenceId={ReferenceId}, EvalId={EvalId}",
-                command.ReferenceId, command.EvalId);
+                SanitizeForLogging(command.ReferenceId), SanitizeForLogging(command.EvalId));
             // Return success to prevent Socure retries — challenge may have been cleaned up
             return Result.Success();
         }
@@ -64,7 +67,7 @@ public class ProcessWebhookCommandHandler(
         {
             logger.LogInformation(
                 "Webhook event {EventId} already processed for challenge {ChallengeId}",
-                command.EventId, challenge.PublicId);
+                SanitizeForLogging(command.EventId), challenge.PublicId);
             return Result.Success();
         }
 
@@ -73,7 +76,7 @@ public class ProcessWebhookCommandHandler(
         {
             logger.LogWarning(
                 "Webhook event {EventId} arrived for terminal challenge {ChallengeId} (status: {Status})",
-                command.EventId, challenge.PublicId, challenge.Status);
+                SanitizeForLogging(command.EventId), challenge.PublicId, challenge.Status);
             return Result.Success();
         }
 
@@ -82,7 +85,7 @@ public class ProcessWebhookCommandHandler(
         {
             logger.LogInformation(
                 "Webhook event {EventId}: intermediate decision {Decision}, challenge {ChallengeId} stays Pending",
-                command.EventId, command.DocumentDecision, challenge.PublicId);
+                SanitizeForLogging(command.EventId), SanitizeForLogging(command.DocumentDecision), challenge.PublicId);
             return Result.Success();
         }
 
@@ -91,7 +94,7 @@ public class ProcessWebhookCommandHandler(
         {
             logger.LogWarning(
                 "Webhook event {EventId} has unrecognized document decision: {Decision}",
-                command.EventId, command.DocumentDecision);
+                SanitizeForLogging(command.EventId), SanitizeForLogging(command.DocumentDecision));
             return Result.Success();
         }
 
@@ -124,13 +127,13 @@ public class ProcessWebhookCommandHandler(
             logger.LogInformation(
                 "Webhook event {EventId}: concurrency conflict on challenge {ChallengeId}, " +
                 "another thread already processed it",
-                command.EventId, challenge.PublicId);
+                SanitizeForLogging(command.EventId), challenge.PublicId);
             return Result.Success();
         }
 
         logger.LogInformation(
             "Webhook event {EventId}: challenge {ChallengeId} transitioned to {Status}",
-            command.EventId, challenge.PublicId, newStatus);
+            SanitizeForLogging(command.EventId), challenge.PublicId, newStatus);
 
         // If verified: update user's proofing status and IAL level
         if (newStatus == DocVerificationStatus.Verified)


src/SEBT.Portal.UseCases/IdProofing/ProcessWebhook/ProcessWebhookCommandHandler.cs
```diff
@@ -0,0 +1,193 @@
...
+        // Fallback lookup by EvalId (D6)
+        if (!string.IsNullOrWhiteSpace(evalId))
+        {
+            return await challengeRepository.GetByEvalIdAsync(evalId, cancellationToken);
+        }
+
+        return null;
+    }
+
+    private static DocVerificationStatus? MapDecisionToStatus(string? decision)
```
noxmwalsh: There's four potential `decision` response according to the Socure Documentation; I think it might be worth covering them.

Image

src/SEBT.Portal.Infrastructure/Repositories/DatabaseDocVerificationChallengeRepository.cs
```diff
@@ -0,0 +1,163 @@
...
+        int userId,
+        CancellationToken cancellationToken = default)
+    {
+        // Non-terminal statuses: Created (0) and Pending (1)
+        var entity = await dbContext.DocVerificationChallenges
+            .AsNoTracking()
+            .FirstOrDefaultAsync(
+                c => c.UserId == userId
+                     && (c.Status == (int)DocVerificationStatus.Created
+                         || c.Status == (int)DocVerificationStatus.Pending),
```
noxmwalsh: We'd probably want to filter out challenges that have expired (for example, if they trigger a challenge, but walk away or close their browser to come back later).  Expanding this out to also filter by `(c.ExpiresAt == null || c.ExpiresAt > DateTime.UtcNow)` would probably cover that.  LMKWYT!

PR #145, opened 2026-04-08
DC-195: (CO) Updating Mailing Address

noxmwalsh:
This is a supporting PR for: https://github.com/codeforamerica/sebt-self-service-portal-co-connector/pull/11

This PR does the following:

* Updates the Update Address Page to include state codes as the given value instead of the fully qualified name
Screenshot 2026-04-08 at 9 38 15 AM
* Hardens state plugin loading so Colorado (and other) connector DLLs register reliably: `PluginAssemblyLoader` resolves plugins-* folders from both the API output directory and the host content root and only skips assemblies already loaded in the default load context (instead of treating every bin DLL as loaded), and keeps dependency resolution from plugin paths. 
* Removes or modifies some logging we want to deal with anyways


src/SEBT.Portal.Infrastructure/Repositories/HouseholdRepository.cs
```diff
@@ -85,17 +75,9 @@ public HouseholdRepository(
 
         if (pluginHousehold == null)
         {
-            _logger.LogInformation(
```
spokenbird, line 88: I think this and the log below we should likely keep for better observability?

src/SEBT.Portal.Infrastructure/Services/HouseholdIdentifierResolver.cs
```diff
@@ -70,16 +70,11 @@ public HouseholdIdentifierResolver(
...
-        _logger?.LogInformation(
-            "Resolving household identifier; preferred types: [{Types}]",
-            string.Join(", ", preferredTypes.Select(t => t.ToString())));
-
         if (preferredTypes.Contains(PreferredHouseholdIdType.Phone))
         {
             var overridePhone = _phoneOverrideProvider.GetOverridePhone();
             if (!string.IsNullOrWhiteSpace(overridePhone))
             {
-                _logger?.LogInformation("Using development phone override for household lookup ");
```
spokenbird, line 82: Also probably good to keep?

src/SEBT.Portal.Web/src/features/address/components/AddressForm/AddressForm.tsx
```diff
@@ -27,16 +27,26 @@ interface FieldErrors {
...
 const STATE_DEFAULTS: Record<string, { city: string; state: string }> = {
-  dc: { city: 'Washington', state: 'District of Columbia' },
-  co: { city: '', state: 'Colorado' }
+  dc: { city: 'Washington', state: 'DC' },
+  co: { city: '', state: 'CO' }
 }
 
-/** Resolves a state value to a dropdown option: tries exact match, then abbreviation lookup, then fallback. */
+/** Resolves backend/form state to the USPS code used as the select value (labels stay full names). */
 function resolveStateValue(value: string | null | undefined, fallback: string): string {
```
spokenbird, line 35: Is the double pass here for differences between CO and DC?

src/SEBT.Portal.Api/Composition/PluginAssemblyLoader.cs
```diff
@@ -95,4 +103,61 @@ public static List<Assembly> LoadAssembliesFromPaths(
...
+                if (!Directory.Exists(dir))
+                    continue;
+                try
+                {
+                    if (Directory.GetFiles(dir, "*.dll", searchOption).Length == 0)
+                        continue;
+                }
+                catch (IOException)
+                {
+                    continue;
```
spokenbird, line 151: Nit: maybe a log warn here?

spokenbird (approved):
Maybe keep some of the logging? But non blocking

src/SEBT.Portal.Infrastructure/Repositories/HouseholdRepository.cs
noxmwalsh: I can add all these back.  These were logs that were previously just use to debug certain issues locally, but I can restore these back.

PR #146, opened 2026-04-08
DC-260 Setup rollment checker in AWS with tofu IaC & CI/CD pipeline

spokenbird:


necampanini (changes requested):
conceptually this is great, just one potential blocking issue. the proxied call from the browser to the .net controller may fail since it only has `[HttpPost("check")]`, and a CORS call (i believe) fires an `OPTIONS` request first

necampanini (approved):
looks good! latest commit addressed the blocking preflight issue

PR #155, opened 2026-04-10
DC-271 Add OTP bypass feature for DAST scanning

jhernandez-cfa:
### Summary

Adds an OTP bypass mechanism for DAST (Dynamic Application Security Testing) scanning in staging. The bypass allows automated security scanning to authenticate without a real OTP, gated behind a triple-guard:

1. **Feature flag** `bypass_otp` must be enabled
2. **Environment** must be `Staging`
3. **Email** must match the hardcoded DAST test account `dast-sanner@sebtportal.com`
4. **OTP code** (validate only) must match `123456`

The bypass logic lives in the **use case handlers** (`RequestOtpCommandHandler`, `ValidateOtpCommandHandler`) — not in the controller. This keeps the controller as a thin HTTP adapter and ensures the bypass is testable at the unit level without HTTP concerns.

#### Production code

| File | Change |
|---|---|
| RequestOtpCommandHandler.cs | Injected `IFeatureManager` + `IHostEnvironment`. Added bypass check at top of `Handle()` — returns `Result.Success()` immediately without generating, persisting, or sending an OTP. |
| ValidateOtpCommandHandler.cs | Same injection. Bypass check additionally requires `command.Otp == "123456"`. On bypass, produces a **real JWT** by calling `GetOrCreateUserAsync` → IAL1 upgrade → `GenerateToken`. No OTP repository lookup or deletion. |
| SEBT.Portal.UseCases.csproj | Added `Microsoft.FeatureManagement` and `Microsoft.Extensions.Hosting.Abstractions` package references. |
| `appsettings.json` | Added `FeatureManagement` configuration section with the `bypass_otp` flag. |

#### Test code

| File | Change |
|---|---|
| RequestOtpCommandHandlerTests.cs | Updated handler setup to include new dependencies. Added **4 bypass tests**: bypass fires correctly, does not fire in non-staging, does not fire for wrong email, does not fire when flag is disabled. |
| ValidateOtpCommandHandlerTests.cs | Updated all 18 handler instantiations with new dependencies. Added **5 bypass tests**: bypass produces real JWT, does not fire in non-staging, does not fire for wrong email, does not fire for wrong OTP, does not fire when flag is disabled. |
| `SEBT.Portal.Tests.csproj` | Updated `Microsoft.FeatureManagement` from 3.3.0 → 4.4.0 to align with the UseCases project dependency. |

#### What did NOT change

The **controller** (`OtpController.cs`) has no bypass-related code — no `IFeatureManager`, no `IHostEnvironment`, no conditional logic. It remains unchanged from `main`.

### Design decisions

1. **Bypass logic in handlers, not controllers** — The handlers own the business logic; the controller is just HTTP plumbing. This makes the bypass fully testable via unit tests without needing HTTP context.

2. **Real JWT on bypass** — The validate bypass calls `GetOrCreateUserAsync` + `GenerateToken` to produce a valid JWT. This allows the DAST scanner to continue testing authenticated flows (household lookup, address update, etc.) with a real token.

3. **Hardcoded email and OTP** — The bypass email (`dast-sanner@sebtportal.com`) and OTP (`123456`) are intentionally hardcoded constants rather than configuration values. Externalizing them would make it easier to accidentally expand the bypass surface.

4. **Triple/quadruple guard** — Feature flag AND staging environment AND email match (AND OTP match for validate). All conditions must be true. Any single failure causes the request to proceed through the normal OTP flow.

### Testing

- **47 tests pass** (13 existing + 4 new for RequestOtp, 23 existing + 5 new for ValidateOtp, 12 existing for OtpController)
- Bypass tests verify both the happy path (bypass fires) and negative paths (each guard condition independently prevents bypass)
- Existing tests are unaffected — the mocked `IFeatureManager` returns `false` by default, so bypass never fires in existing test cases

### How to verify

```bash
dotnet build
dotnet test --filter "FullyQualifiedName~RequestOtpCommandHandler|FullyQualifiedName~ValidateOtpCommandHandler|FullyQualifiedName~OtpControllerTests"
```


src/SEBT.Portal.UseCases/Auth/RequestOtp/RequestOtpCommandHandler.cs
```diff
@@ -20,16 +22,31 @@ namespace SEBT.Portal.UseCases.Auth
...
         : ICommandHandler<RequestOtpCommand>
     {
         public async Task<Result> Handle(RequestOtpCommand command, CancellationToken cancellationToken)
         {
+            // Bypass OTP request for specific email in staging environment when all criteria are met
+            // This allows testing of the login flow without needing to receive an OTP, but only for a specific test email and only in staging
+            if (await featureManager.IsEnabledAsync("bypass_otp")
+                && hostEnvironment.IsStaging()
+                && !string.IsNullOrEmpty(command.Email)
+                && command.Email == "dast-sanner@sebtportal.com")
```
necampanini: is `dast-sanner` here a typo? may silently fail

jhernandez-cfa: Yes, typo and have extracted to class prop to avoid so many strings

test/SEBT.Portal.Tests/Unit/UseCases/Auth/RequestOtpCommandHandlerTests.cs
```diff
@@ -212,4 +218,73 @@ public async Task Handle_ShouldReturnFailure_WhenEmailIsInvalid()
...
         Assert.Contains("Invalid email format.", failedResult.Errors.Select(e => e.Message));
     }
+
+    [Fact]
+    public async Task Handle_WhenBypassEnabled_AndStaging_AndMatchingEmail_ReturnsSuccessWithoutSendingOtp()
+    {
+        // Arrange
+        featureManager.IsEnabledAsync("bypass_otp").Returns(true);
+        hostEnvironment.EnvironmentName.Returns("Staging");
+        var command = new RequestOtpCommand { Email = "dast-sanner@sebtportal.com" };
```
necampanini: `dast-sanner` appears in a few places as well, just making sure if its intentional or not

src/SEBT.Portal.UseCases/Auth/RequestOtp/RequestOtpCommandHandler.cs
```diff
@@ -20,16 +22,31 @@ namespace SEBT.Portal.UseCases.Auth
...
+            {
+                logger.LogWarning("OTP bypass is enabled. Skipping OTP request for email {Email}", command.Email);
```
necampanini: since we're being careful around PII, we should probably not log a user's email here

src/SEBT.Portal.UseCases/Auth/ValidateOtp/ValidateOtpCommandHandler.cs
```diff
@@ -20,16 +22,45 @@ namespace SEBT.Portal.UseCases.Auth
...
     {
         public async Task<Result<string>> Handle(ValidateOtpCommand command, CancellationToken cancellationToken = default)
         {
+            // Bypass OTP validation for specific email in staging environment when all criteria are met
+            // 1. Feature flag "bypass_otp" is enabled
+            // 2. The application is running in the staging environment
+            // 3. The email in the request matches "dast-sanner@sebtportal.com"
+            // 4. The OTP code in the request matches "123456"
+            if (await featureManager.IsEnabledAsync("bypass_otp")
+                && hostEnvironment.IsStaging()
```
necampanini: the other path to this wraps GetOrCreateUserAsync, UpdateUserAsync, and GenerateToken in a try/catch, should we do that here as well?

src/SEBT.Portal.UseCases/Auth/RequestOtp/RequestOtpCommandHandler.cs
```diff
@@ -20,16 +22,31 @@ namespace SEBT.Portal.UseCases.Auth
...
+        ILogger<RequestOtpCommandHandler> logger,
+        IFeatureManager featureManager,
+        IHostEnvironment hostEnvironment)
         : ICommandHandler<RequestOtpCommand>
     {
         public async Task<Result> Handle(RequestOtpCommand command, CancellationToken cancellationToken)
         {
+            // Bypass OTP request for specific email in staging environment when all criteria are met
+            // This allows testing of the login flow without needing to receive an OTP, but only for a specific test email and only in staging
+            if (await featureManager.IsEnabledAsync("bypass_otp")
```
necampanini: `bypass_otp` appears as a string in a number of places. might be worth make pulling up to a single variable

src/SEBT.Portal.UseCases/SEBT.Portal.UseCases.csproj
```diff
@@ -17,6 +17,11 @@
       <ProjectReference Include="..\SEBT.Portal.Core\SEBT.Portal.Core.csproj" />
     </ItemGroup>
 
+    <ItemGroup>
+      <PackageReference Include="Microsoft.Extensions.Hosting.Abstractions" Version="10.0.5" />
+      <PackageReference Include="Microsoft.FeatureManagement" Version="4.4.0" />
+    </ItemGroup>
```
necampanini, line 21: `UseCases.csproj` and `Tests.csproj` reference FeatureManagement 4.4.0, but `Infrastructure.csproj` uses 3.3.0. 

i think NuGet resolves to the latest so it's not breaking, but might be worth noting

src/SEBT.Portal.UseCases/SEBT.Portal.UseCases.csproj
```diff
@@ -17,6 +17,11 @@
       <ProjectReference Include="..\SEBT.Portal.Core\SEBT.Portal.Core.csproj" />
     </ItemGroup>
 
+    <ItemGroup>
+      <PackageReference Include="Microsoft.Extensions.Hosting.Abstractions" Version="10.0.5" />
```
necampanini: more of a clean-architecture concern than anything, but pulling IHostEnvironment into the UseCases layer adds dependency to the app layer

non-blocking at this point, but we might consider extracting the bypass logic if it grows

src/SEBT.Portal.UseCases/Auth/RequestOtp/RequestOtpCommandHandler.cs
jhernandez-cfa: Agreed, creating an appsettings class to centralize all strings

necampanini (approved):
Changes look good! Not sure how much we're going off the Github Security Bot recommendations, but I assume this is feature is only for a very specific circumstance, so approved on my end

src/SEBT.Portal.Api/Controllers/Auth/OtpController.cs
```diff
@@ -70,15 +84,27 @@ public async Task<IActionResult> RequestOtp(
...
+        if (request == null)
         {
             return BadRequest(new ErrorResponse("Request body is required."));
         }
 
-        logger.LogInformation("OTP validation request received for email {Email}", command.Email);
+        logger.LogInformation("OTP validation request received for email {Email}", request.Email);
+
+        var isOtpByPassEnabled = await featureManager.IsEnabledAsync(OtpBypassSettings.FeatureFlagName)
+                && hostEnvironment.IsStaging()
```
necampanini, line 102: the PR description says the bypassed OTP should be "123456", but that's not referenced here, unsure if intentional or not.

src/SEBT.Portal.UseCases/Auth/ValidateOtp/ValidateOtpCommand.cs
```diff
@@ -18,5 +18,11 @@ public class ValidateOtpCommand : ICommand<string>
         [Required(ErrorMessage = "One time password is required.")]
         [Length(6, 6, ErrorMessage = "One time password must be exactly six digits.")]
         public string Otp { get; init; } = string.Empty;
+
+        /// <summary>
+        /// Indicates whether OTP validation should be bypassed.
+        /// </summary>
+        public bool BypassOtp { get; init; } = false;
+
```
necampanini, line 26: this bool prop would mean the handler trusts whatever the caller sets. currently only OtpController sets it, and only after checking flag + staging + email

any future caller (another controller, new tests, an integration point) could bypass OTP entirely just by flipping this flag. the PR description says the bypass logic lives in the handlers, which would address this and be the overall safer approach. might be worth pulling the guard into the handler so the trust boundary matches the description?

src/SEBT.Portal.UseCases/Auth/RequestOtp/RequestOtpCommand.cs
```diff
@@ -13,4 +13,8 @@ public class RequestOtpCommand : ICommand
     [Required(ErrorMessage = "Email address is required.")]
     [EmailAddress(ErrorMessage = "Invalid email format.")]
     public string Email { get; init; } = string.Empty;
+    /// <summary>
+    /// Indicates whether to bypass OTP generation and sending, allowing for testing of the login flow without needing to receive an OTP. 
+    /// </summary>
+    public bool BypassOtp { get; init; } = false;
```
necampanini, line 19: same concern as the `ValidateOtpCommand` prop

necampanini (changes requested):

necampanini (approved):

PR #162, opened 2026-04-13
DC-273: Fix dashboard error caused by empty card status from CBMS

jamesmblair:
## Summary

- **CardStatusSchema** Zod preprocess now maps empty strings to `'Unknown'` instead of passing them through to fail enum validation. CBMS returns `""` for `ebtCardSts` on children without cards (denied/pending), which broke the dashboard.
- Added **console.error logging** in `apiFetch` for schema validation failures to aid future debugging.
- Suppressed verbose **USWDS compile notifications** (`$theme-show-notifications: false`) in the SASS token generator.

**Companion PR:** See codeforamerica/sebt-self-service-portal-co-connector#17 — the backend fix routes `EbtCardStatus` through `MapCardStatus()` instead of passing the raw CBMS string.

### DIFF
--- a/packages/design-system/design/scripts/generate-sass-tokens.js
+++ b/packages/design-system/design/scripts/generate-sass-tokens.js
@@ -265,6 +265,11 @@ function generateSettingsContent(state, variables, timestamp) {
   settingsLines.push('  $theme-alert-icon-size: 3,')
   settingsLines.push('')
 
+  // Suppress USWDS release notes printed on every compile
+  settingsLines.push('  // Suppress verbose USWDS compile notifications')
+  settingsLines.push('  $theme-show-notifications: false,')
+  settingsLines.push('')
+
   // Add utility settings - ensures utility classes override component styles
   settingsLines.push('  // Utility settings')
   settingsLines.push('  $utilities-use-important: true')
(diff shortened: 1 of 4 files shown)


packages/design-system/design/scripts/generate-sass-tokens.js
```diff
@@ -265,6 +265,11 @@ function generateSettingsContent(state, variables, timestamp) {
   settingsLines.push('  $theme-alert-icon-size: 3,')
   settingsLines.push('')
 
+  // Suppress USWDS release notes printed on every compile
```
jamesmblair, line 268: Verbose USWDS notifications were making it challenging to finding meaningful information in server-side web logs. This disables that noise.

src/SEBT.Portal.Web/src/api/client.ts
```diff
@@ -106,6 +106,8 @@ export async function apiFetch<T>(endpoint: string, options: ApiFetchOptions<T>
   if (schema && data !== undefined) {
     const result = schema.safeParse(data)
     if (!result.success) {
+      console.error(`[apiFetch] Validation failed for ${endpoint}:`, result.error.issues)
```
jamesmblair, line 109: The thrown exception was being caught and suppressed (with an error message in the UI) further up the call stack. This ensures we log meaningful information to the console if schema validation fails.

spokenbird (approved):
Looks good!

PR #163, opened 2026-04-13
DC-276 Fix: Duplicate Row error on DocVerificationChallenges table

noxmwalsh:
This PR adds the following: 

* Before inserting a new Doc Verification challenge, expire any created/pending rows for the same user with non-null (`ExpiresAt` ≤ `UtcNow`), so they no longer satisfy the partial unique index.
* Unit test in `DatabaseDocVerificationChallengeRepositoryTests` added to cover this scenario

### DIFF
--- a/src/SEBT.Portal.Infrastructure/Repositories/DatabaseDocVerificationChallengeRepository.cs
+++ b/src/SEBT.Portal.Infrastructure/Repositories/DatabaseDocVerificationChallengeRepository.cs
@@ -86,6 +86,20 @@ public class DatabaseDocVerificationChallengeRepository(PortalDbContext dbContex
             throw new ArgumentNullException(nameof(challenge));
         }
 
+        var now = DateTime.UtcNow;
+
+        await dbContext.DocVerificationChallenges
+            .Where(c => c.UserId == challenge.UserId
+                && (c.Status == (int)DocVerificationStatus.Created
+                    || c.Status == (int)DocVerificationStatus.Pending)
+                && c.ExpiresAt != null
+                && c.ExpiresAt <= now)
+            .ExecuteUpdateAsync(
+                setters => setters
+                    .SetProperty(e => e.Status, (int)DocVerificationStatus.Expired)
+                    .SetProperty(e => e.UpdatedAt, now),
+                cancellationToken);
+
         var entity = MapToEntity(challenge);
         dbContext.DocVerificationChallenges.Add(entity);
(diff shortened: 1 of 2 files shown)


src/SEBT.Portal.Infrastructure/Repositories/DatabaseDocVerificationChallengeRepository.cs
```diff
@@ -86,6 +86,20 @@ public async Task CreateAsync(
...
+                && c.ExpiresAt <= now)
+            .ExecuteUpdateAsync(
+                setters => setters
+                    .SetProperty(e => e.Status, (int)DocVerificationStatus.Expired)
+                    .SetProperty(e => e.UpdatedAt, now),
+                cancellationToken);
+
         var entity = MapToEntity(challenge);
         dbContext.DocVerificationChallenges.Add(entity);
         await dbContext.SaveChangesAsync(cancellationToken);
```
necampanini: I'd wrap this and the newer block in a transaction to be safe, as `.ExecuteUpdateAsync` and `.SaveChangesAsync` commit independently, and if somethings breaks midway, it could leave the data in a weird state for retries

src/SEBT.Portal.Infrastructure/Repositories/DatabaseDocVerificationChallengeRepository.cs
```diff
@@ -86,6 +86,20 @@ public async Task CreateAsync(
             throw new ArgumentNullException(nameof(challenge));
         }
 
+        var now = DateTime.UtcNow;
+
+        await dbContext.DocVerificationChallenges
+            .Where(c => c.UserId == challenge.UserId
```
necampanini: this works and fixes the immediate issue, my only concern is that it goes around the row-version concurrency & state machine we have.

nothing major, but maybe a comment saying as such and to revisit if necessary

necampanini (changes requested):
looks good and unblocks testing, just small change/comment requested

src/SEBT.Portal.Infrastructure/Repositories/DatabaseDocVerificationChallengeRepository.cs
noxmwalsh: I don't think it would be a big lift to wrap it in a transaction; I'll do that.

necampanini (approved):

PR #164, opened 2026-04-13
DC-278: Fix CO footer color

jamesmblair:
- Footer color was previously hard-coded to always use DC's teal color
- Set state-specific footer color classes
- update footer components to use state-specific footer classes

### DIFF
--- a/packages/design-system/src/components/layout/Footer.tsx
+++ b/packages/design-system/src/components/layout/Footer.tsx
@@ -1,32 +1,30 @@
-"use client";
+'use client'
 
-import Image from "next/image";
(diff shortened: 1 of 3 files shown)


packages/design-system/src/components/layout/Footer.tsx
```diff
@@ -137,5 +136,5 @@ function COFooter({ state = 'co' }: FooterProps) {
         </div>
       </div>
     </footer>
-  )
+  );
```
ShaynaCummings: hrmmm do we like semicolons?

ShaynaCummings (approved):

PR #167, opened 2026-04-14
DC-187: Fix Spanish translations not rendering in client components

ShaynaCummings:
#### 🔗 Jira ticket
[DC-187
](https://codeforamerica.atlassian.net/browse/DC-187) 

#### ✍️ Description

**Root cause**

`COLoginPage` and `CallbackPage` are both `'use client'` components, but were calling `getTranslations()` — a server-side utility hardcoded to load only English resources at module init time. Language switching happens entirely client-side via i18next: when a user selects Spanish, only components using `useTranslation()` from `react-i18next` re-render. Components using `getTranslations()` always return English regardless of the selected language.

**Fix**

Replaced `getTranslations()` with `useTranslation()` in the two affected client components:

- `COLoginPage.tsx` — primary CO login page, where the bug was first reported
- `callback/page.tsx` — OIDC callback page (same pattern, same bug)

One nuance in `CallbackPage`: `t` is intentionally omitted from the `useEffect` dependency array to avoid re-triggering the OIDC token exchange on language change. The eslint-disable comment was updated to explain the real reason.

**Out of scope / follow-up**

- `HelpSection.tsx` (`COHelpSection`) has hardcoded English strings with TODO comments — those are a content gap (keys not yet in `co.csv`) blocked on the content team updating the Google Sheet.
- `CallbackPage` translation keys (`callbackSigningIn`, `callbackErrorGeneric`, etc.) do not exist in any locale CSV yet — noted as a content gap in the test file.

#### ✅ Completion tasks

- [x] Added relevant tests
- [ ] Configuration changes:

---

#### Test changes

**`COLoginPage.test.tsx`** *(new)*
Added a test suite for `COLoginPage`, including:
- English rendering assertions (title, buttons)
- A language-switch regression test: renders in English, calls `i18n.changeLanguage('es')`, and asserts the Spanish heading appears. This test fails if the component is reverted to `getTranslations()`, since that function ignores `changeLanguage`.

**`login/page.test.tsx`**
The CO tests previously relied on a mock for `@/lib/translations` to supply translation values. Since `COLoginPage` now uses `useTranslation`, the CO `describe` block now injects real CO locale bundles (`en/co/login.json`, `en/co/common.json`) into the live i18n instance via `addResourceBundle`. Three assertions were corrected to match actual locale values (they had been written against the old mock's invented strings):
- Title: `"Access your Summer EBT account"` → `"Access the Summer EBT portal"`
- Esp button: `Iniciar sesión con myColorado` → `Iniciar sesión myColorado` ("con" is not in the CO locale)
- Contact link: `"...logging into your account."` → `"...logging in."`

**`callback/page.test.tsx`**
Removed the `@/lib/translations` mock (no longer imported by the component). The callback-specific translation keys do not exist in the locale CSV files yet, so they are injected as a test-only bundle via `addResourceBundle` with a comment flagging them as a content gap.


necampanini (approved):

src/SEBT.Portal.Web/src/app/(public)/login/COLoginPage.test.tsx
```diff
@@ -0,0 +1,102 @@
+/**
+ * COLoginPage Unit Tests
+ *
+ * Includes a regression test for language switching:
```
necampanini: I think the other pages that the `useTranslation()` fix was applied to could use similar regression test as well to keep it consistent

src/SEBT.Portal.Web/src/app/(public)/callback/page.tsx
necampanini: the test injects the missing keys, which I think is fine, but the frontend had been hardcoding `TODO`s for using a key once its available. as is, key names are making it through to the UI. 

not introduced by this PR, just noting it

necampanini, line 45: could be wrong but I think that error text would get locked to whatever language they were in when thrown. so storing the key instead of the resolved string would let it be reactive

```
useEffect(() => {
    if (!code || !state) {
      setErrorKey('callbackErrorMissingParams')  // store the key
      setStep('error')
      setStatus('error')
      return
    }
    // ...
  }, [login, router])
```

src/SEBT.Portal.Web/src/app/(public)/login/page.test.tsx
necampanini, line 1: running these tests locally are failing, which i think has more to do with being behind in general

src/SEBT.Portal.Web/src/app/(public)/callback/page.tsx
necampanini: still a potential issue after newer branch. error text would get locked to whatever language it was thrown in, despite clicking the language selection

necampanini (changes requested):
looking good, just the error text getting frozen in language issue remaining

necampanini (approved):

PR #173, opened 2026-04-15
DC-291: Co-loaded case filtering and per-case feature flags

jamesmblair:
## Summary

- Filter co-loaded cases (SNAP/TANF benefits on existing EBT cards) from the `GetHouseholdData` API response so portal users only see non-co-loaded cases
- Replace `BenefitIssuanceType` checks with `IsCoLoaded` in `UpdateAddressCommandHandler` and add co-loaded guard to `RequestCardReplacementCommandHandler`
- Add per-case `AllowAddressChange` and `AllowCardReplacement` flags to the API response, derived from `!IsCoLoaded`
- Gate the "Change my mailing address" link and address form page on the new flags
- Fix mock data: add DC mailing address to Review scenario, mark Verified scenario cases as co-loaded
- Add TDD for future Smarty address autocomplete integration

## Test plan

- [ ] Backend: 869 tests passing (4 new tests for co-loaded filtering/rejection, 4 replacement tests for IsCoLoaded-based address update checks)
- [ ] Frontend: 556 tests passing (2 new tests for address link visibility gating)
- [ ] Manual: Log in as `sebt.dc+verified` (co-loaded) — verify address change link is hidden and `/profile/address` redirects to `/profile`
- [ ] Manual: Log in as `sebt.dc+review` (non-co-loaded) — verify address is displayed and change address link is visible
- [ ] Manual: Log in as `sebt.dc+simple1` (non-co-loaded) — verify address change flow works end-to-end


src/SEBT.Portal.UseCases/Household/UpdateAddress/UpdateAddressCommandHandler.cs
```diff
@@ -128,15 +128,14 @@ public async Task<Result<AddressValidationResult>> Handle(UpdateAddressCommand c
             }
         }
 
-        if (household is { BenefitIssuanceType: BenefitIssuanceType.SnapEbtCard or BenefitIssuanceType.TanfEbtCard })
+        if (household != null && household.SummerEbtCases.Any(c => c.IsCoLoaded))
```
adbergen, line 131: Rejects mixed households that the UI shows as allowed. Is this intended and UI is just wrong/behind?

src/SEBT.Portal.Web/src/app/(authenticated)/profile/address/(flow)/page.tsx
```diff
@@ -1,23 +1,30 @@
...
 // TODO (DC-153): Card-flow entry point — when accessed via /profile/address?from=cards,
 // the form should return the user to the card replacement flow on completion
 // instead of the replacement card prompt. See DC-02/CO-01 mockups.
 
 export default function AddressFormPage() {
   const { t } = useTranslation('confirmInfo')
   const { data, isLoading } = useHouseholdData()
+  const router = useRouter()
+
+  const canChangeAddress = data?.summerEbtCases.some((c) => c.allowAddressChange) ?? false
```
adbergen: canChangeAddress not memoized (booleans are safe but fragile pattern)

adbergen (changes requested):

src/SEBT.Portal.UseCases/Household/UpdateAddress/UpdateAddressCommandHandler.cs
jamesmblair: Yes, for MVP we're not supporting mixed co-loaded/non-co-loaded. Even if the UI technically allows it, the query handler for loading the household should also be applying this filtering logic now.

src/SEBT.Portal.Web/src/app/(authenticated)/profile/address/(flow)/page.tsx
jamesmblair: Addressed by this [commit](https://github.com/codeforamerica/sebt-self-service-portal/pull/173/changes/95a1663d2794dc1cde1ffd3de3225ca3a22d0785)

adbergen (approved):
LGTM!

PR #174, opened 2026-04-15
Add Socure SDK env vars for CO and move web builds to deploy jobs

spokenbird:
- Add `NEXT_PUBLIC_SOCURE_SDK_KEY`, `NEXT_PUBLIC_SOCURE_DI_SDK_KEY`, and      
  `NEXT_PUBLIC_MOCK_SOCURE` as Docker build args so they get baked into the CO  
  web image's client-side JS bundle at build time                               
- Add `NEXT_PUBLIC_MOCK_SOCURE` to the t3-env validation in `env.ts` (the     
  other two already existed)                  
- Move web image builds from the shared `build` job into the per-state        
  `deploy-dc`/`deploy-co` jobs so each build runs with the correct GitHub
  Environment context
  
## Why move the web builds?                                                   
                                          
`NEXT_PUBLIC_*` variables are inlined into the client-side JavaScript bundle  
during `next build`. They must be present as environment variables when       
`docker build` runs. Setting them as ECS task definition env vars at runtime
has no effect on client-side code.                                            
                  
The CO Socure vars come from GitHub Environment variables (`dev-co`), which   
are only available to jobs with `environment: dev-co`. The shared `build` job
had no environment set, so it couldn't access them. Moving each state's web
build into its deploy job gives it the right environment context.             
  
A side benefit: DC and CO web images now build in parallel (one per deploy    
job) instead of sequentially, and the shared artifact is smaller (API image
only).
  
## Test plan                                                                  
                  
- [x] Verify `dev-co` GitHub Environment variables are set before merging     
- [ ] Confirm deploy-ecr workflow completes successfully for both DC and CO
  after merge                                                                   
- [ ] Verify Socure SDK initializes correctly in the CO web app               
- [ ] Verify DC deployment is unaffected (no Socure vars, builds as before)

### DIFF
--- a/.github/workflows/deploy-ecr.yaml
+++ b/.github/workflows/deploy-ecr.yaml
@@ -187,30 +187,10 @@ jobs:
             -t sebt-portal-api:${{ steps.tag.outputs.tag }} \
             .
 
-      - name: Build Web DC image
-        run: |
-          docker build \
-            --platform linux/amd64 \
-            -f src/SEBT.Portal.Web/Dockerfile \
-            --build-arg STATE=dc \
-            -t sebt-portal-web-dc:${{ steps.tag.outputs.tag }} \
-            .
-
-      - name: Build Web CO image
-        run: |
-          docker build \
-            --platform linux/amd64 \
-            -f src/SEBT.Portal.Web/Dockerfile \
-            --build-arg STATE=co \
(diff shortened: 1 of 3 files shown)


noxmwalsh (approved):
I've been able to build the web Docker image fine with the new args.  Approved

PR #177, opened 2026-04-16
DC-294 Fix: MyCO sign-in and step-up use correct language

jamesmblair:
## Summary

- Spanish MyCO login button now redirects to Spanish MyCO instead of English
- Step-up identity verification redirects to MyCO in the user's current UI language
- Portal returns in the correct language after MyCO redirects back
- Removed `Oidc:LanguageParam` from backend config — language is now a frontend concern

## What was wrong

The OIDC `language` query parameter was read from backend config (`Oidc:LanguageParam`), which was hardcoded to `"en"`. Both the login page and step-up flow passed this value through unchanged, so Spanish-language users always landed on English MyCO.

Additionally, when clicking the Spanish login button, the language choice wasn't persisted to `localStorage` before the redirect, so the portal initialized in English after MyCO redirected back.

## Test plan

- [ ] Click Spanish login button → MyCO loads in Spanish
- [ ] Click English login button → MyCO loads in English
- [ ] Complete Spanish MyCO login → portal loads in Spanish
- [ ] Trigger step-up verification while in Spanish UI → MyCO loads in Spanish
- [ ] `pnpm test` passes (oidc-pkce tests updated for new `language` parameter)


ShaynaCummings (approved):
LGTM

PR #180, opened 2026-04-16
DC-283 - Server-side OIDC authorize redirect (V04 security fix)

spokenbird:
#### 🔗 Jira ticket
DC-283

#### ✍️ Description

Fixes the remaining V04 vulnerability (config hijacking / credential phishing) identified in the CO security re-evaluation. The `/api/auth/oidc/{code}/config` endpoint previously returned the PingOne `authorizationEndpoint` URL to the browser, allowing a MitM attacker to tamper the response and redirect users to a fake login page.

**What changed:**
- New `GET /api/auth/oidc/{code}/authorize` endpoint builds the full authorization URL server-side and returns a **302 redirect**. The browser follows the redirect directly to PingOne — it never sees or handles the authorization endpoint URL.
- The `authorization_endpoint` is now read from the OIDC discovery document at runtime (standard OIDC approach) instead of being pinned in `Oidc:AuthorizationEndpoint` config. That config key has been removed.
- The `/config` endpoint no longer returns `authorizationEndpoint` in its response.
- `Callback` and `CompleteLogin` endpoints now read `stateCode`, `isStepUp`, and `returnUrl` from the server-side pre-auth session instead of the request body — the session is authoritative for values set at authorize time.
- Frontend (`COLoginPage`, `IalGuard`, callback page) navigates directly to the `/authorize` endpoint instead of fetching config and building the URL client-side. `sessionStorage` is no longer used for OIDC flow metadata.
- `oidc-pkce.ts` (client-side URL building and sessionStorage helpers) has been removed.
- Next.js API proxy now passes backend 302 redirects through to the browser (`redirect: 'manual'`).

**Commits:**
1. Add `ReturnUrl` to `PreAuthSession` — server-side storage for step-up return paths
2. Extract discovery document fetching into shared service on `OidcExchangeService`
3. New `Authorize` endpoint + remove `Oidc:AuthorizationEndpoint` from config
4. Strip `authorizationEndpoint` from `GetConfig`, update `Callback`/`CompleteLogin` to read from session
5. Frontend migration — `COLoginPage`, `IalGuard`, callback page, cleanup

**Security test coverage:**
- Added integration tests for the `Authorize` endpoint's stateCode allowlist validation (replaces removed V05/T07a and V06/T08aE tests that validated body-level stateCode, which is no longer read)
- Unit tests cover: 302 redirect with correct URL params, cookie set, returnUrl sanitization, code_verifier not leaked, discovery failure handling

#### 🔗 Links to related PRs
- #154 — PR that introduced the server-side pre-auth session architecture (merged)

#### ✅ Completion tasks

- [x] Added relevant tests
- [ ] Configuration changes:
  - [x] If you're adding an appsetting, add it to the requisite example file, and update it in the AWS AppConfig — *removed `Oidc:AuthorizationEndpoint` and `Oidc:StepUp:AuthorizationEndpoint` from example appsettings; these keys should also be removed from AppConfig and environment variables in deployed environments*
  - [x] If appsetttings are changed, update in AWS AppConfig — *`Oidc:AuthorizationEndpoint` removed; `StateAllowlist` now derives from `Oidc:DiscoveryEndpoint` instead*


src/SEBT.Portal.Api/Controllers/Auth/OidcController.cs
```diff
@@ -87,25 +84,148 @@ public async Task<IActionResult> GetConfig(
...
+        var clientId = stepUp ? config["Oidc:StepUp:ClientId"] : config["Oidc:ClientId"];
+        var redirectUri = stepUp
+            ? (config["Oidc:StepUp:RedirectUri"] ?? config["Oidc:CallbackRedirectUri"])
+            : config["Oidc:CallbackRedirectUri"];
+        if (string.IsNullOrEmpty(clientId) || string.IsNullOrEmpty(redirectUri))
+        {
+            logger.LogWarning(
+                "OIDC config missing for stateCode {StateCode} (reason=oidc_not_configured)",
+                stateCode);
+            return Redirect("/login");
```
spokenbird, line 152: These error redirects silently land the user on /login with no indication of what went wrong which isn't a great user experience. We could pass a query param (e.g. /login?error=oidc_config) and have COLoginPage show the user something meaningful like "There was an issue handling your request, please try again later".   

Deferred because these error paths are unlikely in practice (config is validated at startup via the StateAllowlist, though we could hit these error paths if there is a PingOne outage?).

src/SEBT.Portal.Web/src/features/auth/components/IalGuard/IalGuard.tsx
```diff
@@ -103,14 +80,10 @@ export function IalGuard({ children, requiredIal = STEP_UP_REQUIRED_IAL }: IalGu
     }
   }, [router])
 
-  const handleVerify = useCallback(async () => {
+  const handleVerify = useCallback(() => {
```
spokenbird, line 83: Note `handleVerify` was previously async, it fetched the OIDC config from the server, and if that fetch failed, it caught the error and showed a friendly in page error screen with "Back" and "Continue" buttons. With the V04 fix, the step-up flow is now a direct browser navigation to `/api/auth/oidc/{stateCode}/authorize`, which returns a 302 redirect. Browser navigations can't be caught by JavaScript, there's no fetch to wrap in a try/catch. So if the server-side authorize endpoint fails (e.g. PingOne is down), the user is silently redirected to `/login` instead of seeing the in page error. The error phase UI (lines 107-146) is still in the component but is currently unreachable. Calling this out because I couldn't think of a good alternative in the time crunch we had to get this ready, but we may want to revisit?

src/SEBT.Portal.Web/src/app/(public)/callback/page.tsx
```diff
@@ -36,14 +36,10 @@ export default function CallbackPage() {
...
+    // IdP returned an error (e.g., user cancelled login).
     if (errorParam) {
-      const storedPkce = getPkceFromStorage()
-      const stepUpFromIdpError = storedPkce?.isStepUp === true
       const idpDetail = errorDescription?.trim() ?? ''
-      const portalLine = t(
-        stepUpFromIdpError ? 'callbackErrorStepUpFailed' : 'callbackErrorIdpRedirect',
-        t('callbackErrorGeneric')
-      )
+      const portalLine = t('callbackErrorIdpRedirect', t('callbackErrorGeneric'))
```
spokenbird, line 42: Note this previously used `sessionStorage` to check `isStepUp` and show a different message for step-up errors (`callbackErrorStepUpFailed`: "Step-up verification did not finish") vs normal login errors (`callbackErrorIdpRedirect`: "Primary MyColorado sign-in did not finish"). With the V04 fix, sessionStorage is no longer used for OIDC flow metadata, all flow state is server-side. Since the IdP error redirect (?error=) happens before the callback POST (PingOne redirects here without a code), we never call the server, so we can't read isStepUp from the session. We now always show the generic message. The copy difference is minor and both convey the same information to the user. We can revisit if we want the explicit distinction?

src/SEBT.Portal.Api/Controllers/Auth/OidcController.cs
mtodd-cfa: maybe just a flag like error=true to avoid exposing unnecessary details, but yeah, agree.

spokenbird: Fair 😅 my brain is pretty fried by the time I got to writing these notes, but agreed, we shouldn't provide any more information than neccessary 👍

src/SEBT.Portal.Web/src/app/api/[[...path]]/route.ts
```diff
@@ -39,6 +39,9 @@ async function proxyRequest(request: NextRequest, context: RouteContext): Promis
       headers,
       body: request.body,
       signal: controller.signal,
+      // Pass backend redirects (e.g., OIDC authorize 302) through to the browser
+      // instead of following them within the proxy.
+      redirect: 'manual',
```
Comment (jamesmblair) [inline src/SEBT.Portal.Web/src/app/api/[[...path]]/route.ts:44]: This changes the redirect behavior globally, right? I don't *think* this should impact other features because I don't think we're redirecting for anything else, but just noting possible future impact.

I don't think this needs to block merging.

jamesmblair (approved):
Approved with one non-blocking comment

jamesmblair (approved):

PR #183, opened 2026-04-17
DC-136(Feat): As a system, I want to capture all DC non co-loaded users datasets available

noxmwalsh:
#### 🔗 Jira ticket
[DC-136](https://codeforamerica.atlassian.net/browse/DC-136)
#### ✍️ Description

Extends the DC connector so household lookups can carry guardian name and phone from the state data source into HouseholdData (`UserProfile` and `Phone`), with SQL/scripts and seeds updated for local dev and the mapping written to tolerate older procs that don’t yet return those columns. Phone on the aggregated household respects `IncludePhone`; mailing address behavior in the DC proc is unchanged (still gated on identity-proofed for street-level fields). We also removed an empty stray SQL file and tightened the test fixture so empty script files fail with a clear message instead of a cryptic SqlClient error.

On the portal side, ID proofing already had (or gained) logic to load household PII when submitting to Socure: names, address when available, and phone with a sensible merge order so the evaluation request can include as much DC-sourced context as the connector returns. Supporting work included Socure HTTP payload adjustments, tests, and a direct System.Security.Cryptography.Xml package reference to clear a security advisory. 

**Production still depends on DC’s real database/proc supplying those guardian fields (and on policy for address before proofing); the portal only forwards what the plugin returns.**

#### 🔗 Links to related PRs
https://github.com/codeforamerica/sebt-self-service-portal-dc-connector/pull/15

#### ✅ Completion tasks

Testing guide:
- Go through the Socure workflow (you have to be a non-co-loaded user; I recommend using the `not-started` seeded user for testing out, but any with an IAL of 0 will work).
- Inspect `api/evaluation` once you get to this screen.
Screenshot 2026-04-14 at 9 23 20 AM

- [ ] Added relevant tests
- [ ] Configuration changes:

[DC-136]: https://codeforamerica.atlassian.net/browse/DC-136?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


src/SEBT.Portal.Infrastructure/SEBT.Portal.Infrastructure.csproj
```diff
@@ -41,6 +41,7 @@
       <PackageReference Include="Microsoft.Extensions.Options.ConfigurationExtensions" Version="10.0.0" />
       <PackageReference Include="Microsoft.FeatureManagement" Version="3.3.0" />
       <PackageReference Include="System.IdentityModel.Tokens.Jwt" Version="8.2.1" />
+      <PackageReference Include="System.Security.Cryptography.Xml" Version="10.0.6" />
```
noxmwalsh: This is to fix a security vulnerability that was reported

src/SEBT.Portal.UseCases/IdProofing/SubmitIdProofing/SubmitIdProofingCommandHandler.cs
```diff
@@ -86,15 +87,16 @@ public async Task<Result<SubmitIdProofingResponse>> Handle(
...
         string? givenName = null;
         string? familyName = null;
         Address? address = null;
+        string? householdPhone = null;
         try
         {
             var household = await householdRepository.GetHouseholdByEmailAsync(
                 user.Email,
-                new PiiVisibility(IncludeAddress: true, IncludeEmail: false, IncludePhone: false),
+                new PiiVisibility(IncludeAddress: true, IncludeEmail: true, IncludePhone: true),
```
necampanini, line 108: I think a comment here is worth documenting the why, since `PiiVisibility` is normally built from the `idProofingRequirementsService` method.

test/SEBT.Portal.Tests/Unit/UseCases/IdProofing/SubmitIdProofingCommandHandlerTests.cs
necampanini, line 1: some tests we could add:
- additional tests to catch reordering of `??` chain?
> 1. Sandbox override set + household has phone + user has phone -> Socure gets the sandbox value.
> 2. No sandbox + household has phone + user has phone -> Socure gets the household phone.
> 3. No sandbox + household phone null/empty + user has phone -> Socure gets `user.Phone`.

- household lookup failure -> Socure still being called could use a test
- asserting `Phone` makes it to `SocureIndividaul.Phone`

necampanini (changes requested):
Overall looks great, just some more tests I'd add

src/SEBT.Portal.UseCases/IdProofing/SubmitIdProofing/SubmitIdProofingCommandHandler.cs
noxmwalsh: Yeah, this is a good point and could get confusing.  There's a bit of a split now where we originally had the thought that we'd never want to get Household/Case data to reach the portal depending on PII rules, but there's cases where we need to have it available to send it elsewhere (in this case, Socure) or to evaluate things within the portal itself (where the client would never see it if they're not at the right IAL).  I'll add a comment explaining this, but I think we need to consider either a refactor or a new interface that'll handle 'server eyes only' fetching for this data where it'll throw if it's ever returned for display purposes.

test/SEBT.Portal.Tests/Unit/UseCases/IdProofing/SubmitIdProofingCommandHandlerTests.cs
noxmwalsh: Addressed in 60ca0954fc02f616a37fa2e84f12f1c2f4155845

src/SEBT.Portal.UseCases/IdProofing/SubmitIdProofing/SubmitIdProofingCommandHandler.cs
noxmwalsh: Addressed in 60ca0954fc02f616a37fa2e84f12f1c2f4155845

noxmwalsh:
@necampanini Ready for re-review!

necampanini (approved):

PR #187, opened 2026-04-20
DC-292: IdP claims as source of truth for IAL

jamesmblair:
## Jira ticket
https://codeforamerica.atlassian.net/browse/DC-292

## Summary

- **Fix stale IAL bug** for OIDC users (Colorado): the portal was reading IAL from its database instead of from the identity provider's claims, causing IAL to go stale between logins
- **Add `ExternalProviderId`** to User model — OIDC users are now identified by IdP `sub` claim instead of email
- **Stop persisting IAL to DB** for OIDC users — IAL, email, and verification state flow from IdP claims at login time and are carried in the JWT
- **In-flight migration** for existing users: legacy email-only records are adopted on next OIDC login by setting `ExternalProviderId` and clearing email
- **Token refresh** is now OIDC-aware: preserves IAL from existing JWT claims instead of overriding with stale DB values

## Known issues to fix before merge

1. `ResolveUserFilter` uses email-only DB lookup — breaks for OIDC users with null email
2. Missing `IdProofingExpiresAt` claim for OIDC users — frontend treats verification as stale
3. Token refresh OIDC detection heuristic (`sub != email`) is fragile — should use explicit claim

## Test plan

- [ ] Run full backend test suite
- [ ] Run frontend test suite
- [ ] Local smoke test: CO OIDC login → verify JWT contains correct IAL
- [ ] Local smoke test: token refresh preserves IAL
- [ ] Local smoke test: DC OTP login still works
- [ ] Verify Users table: CO user has ExternalProviderId, no email; DC user has email, no ExternalProviderId

### DIFF
--- a/docs/superpowers/plans/2026-04-17-oidc-user-claims-source-of-truth.md
+++ b/docs/superpowers/plans/2026-04-17-oidc-user-claims-source-of-truth.md
@@ -1,0 +1,1198 @@
+# OIDC User: IdP Claims as Source of Truth
+
+> **For agentic workers:** REQUIRED SUB-SKILL: Use superpowers:subagent-driven-development (recommended) or superpowers:executing-plans to implement this plan task-by-task. Steps use checkbox (`- [ ]`) syntax for tracking.
+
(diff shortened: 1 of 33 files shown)


ShaynaCummings (approved):
Looks good, appears to be working locally as well now with new and old accounts

PR #196, opened 2026-04-21
DC-318: resolve Smarty HttpClient DI lifetime mismatch

jamesmblair:
## Jira ticket
[DC-318](https://codeforamerica.atlassian.net/browse/DC-318?atlOrigin=eyJpIjoiMGZiNGViYjExYThjNGFiZjk1OWU1ODk4MTcxY2IyYmEiLCJwIjoiaiJ9)

## Description
The `AddHttpClient("Smarty", ...)` configuration delegate in `Dependencies.cs` was resolving `IOptionsSnapshot<SmartySettings>` (scoped) from the root `IServiceProvider`. `IHttpClientFactory` is a singleton, so its configuration delegates always receive the root provider — resolving a scoped service from root throws `InvalidOperationException`.

**Root cause:** `IOptionsSnapshot<T>` is scoped; `AddHttpClient` delegates run in singleton context.

**Fix:** Switch to `IOptionsMonitor<T>.CurrentValue` (singleton, supports live reload).

**Why it worked locally:** In local dev, state overlays (`appsettings.dc.json`) set `Smarty:Enabled = false`, so the `PassThroughAddressUpdateService` was used and the buggy HttpClient delegate never executed. In AWS, Smarty was enabled, triggering the code path.

## Related PRs
None

## Completion checklist
- [x] Tests written and passing (751 unit tests, 0 failures)
- [x] TDD approach: wrote failing test first, then fixed
- [x] No new dependencies added
- [x] Regression test prevents future lifetime mismatches in HttpClient delegates

## Test plan
- [ ] Deploy to dev-dc or dev-co with `Smarty:Enabled = true` and confirm address update no longer throws
- [ ] Verify existing address update behavior unchanged when Smarty is disabled

[DC-318]: https://codeforamerica.atlassian.net/browse/DC-318?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/SEBT.Portal.Infrastructure/Dependencies.cs
+++ b/src/SEBT.Portal.Infrastructure/Dependencies.cs
@@ -52,7 +52,10 @@ public static class Dependencies
         // Smarty address verification (or pass-through when disabled)
         services.AddHttpClient("Smarty", (sp, client) =>
         {
-            var smarty = sp.GetRequiredService<IOptionsSnapshot<SmartySettings>>().Value;
+            // IOptionsMonitor (singleton) instead of IOptionsSnapshot (scoped) — the
+            // AddHttpClient delegate receives the root IServiceProvider, so scoped
+            // services cannot be resolved here.
+            var smarty = sp.GetRequiredService<IOptionsMonitor<SmartySettings>>().CurrentValue;
             var baseUrl = string.IsNullOrWhiteSpace(smarty.BaseUrl)
                 ? "https://us-street.api.smartystreets.com"
                 : smarty.BaseUrl.TrimEnd('/');
--- a/test/SEBT.Portal.Tests/Unit/Infrastructure/DependenciesTests.cs
+++ b/test/SEBT.Portal.Tests/Unit/Infrastructure/DependenciesTests.cs
@@ -32,4 +32,41 @@ public class DependenciesTests
         Assert.Contains("UseMockHouseholdData is false", ex.Message);
         Assert.Contains("no household plugin", ex.Message);
     }
+
+    [Fact]
+    public void CreateSmartyHttpClient_CanBeCreatedFromScope_WhenSmartyEnabled()
+    {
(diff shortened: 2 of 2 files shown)


src/SEBT.Portal.Infrastructure/Dependencies.cs
```diff
@@ -52,7 +52,10 @@ public static IServiceCollection AddPortalInfrastructureServices(this IServiceCo
         // Smarty address verification (or pass-through when disabled)
         services.AddHttpClient("Smarty", (sp, client) =>
         {
-            var smarty = sp.GetRequiredService<IOptionsSnapshot<SmartySettings>>().Value;
+            // IOptionsMonitor (singleton) instead of IOptionsSnapshot (scoped) — the
+            // AddHttpClient delegate receives the root IServiceProvider, so scoped
+            // services cannot be resolved here.
+            var smarty = sp.GetRequiredService<IOptionsMonitor<SmartySettings>>().CurrentValue;
```
ShaynaCummings, line 58: as discussed - nonblocking but would good to add some kind of error handling here so that we could handle smarty configs being wrong / unavailable. Ie, having the ability to gracefully handle so the user can still continue with the address update, even if not validated

ShaynaCummings (approved):
Looks good, one small point as discussed

src/SEBT.Portal.Infrastructure/Dependencies.cs
jamesmblair: Will be handled by [DC-319](https://codeforamerica.atlassian.net/browse/DC-319?atlOrigin=eyJpIjoiNTUwNjk3NjMzYTAyNDcwODg0NTM0ZGRmMTA4YTU3OTYiLCJwIjoiaiJ9)

PR #198, opened 2026-04-22
DC-321(Bug) DocV Transaction token reminting

noxmwalsh:
#### 🔗 Jira ticket
[DC-321](https://codeforamerica.atlassian.net/browse/DC-321)

#### ✍️ Description

**Problem:** Socure DocV **transaction tokens** can become **stale** while the user still has a valid **challenge**. Returning the cached token from `StartChallenge` could strand them in a broken DocV session.

**Approach:**

- **`SocureSettings.DocvTransactionTokenTtlMinutes`** (default **20**, validated **1–120**): after this many minutes from **`DocvTokenIssuedAt`**, the token is treated as stale.
- **`DocVerificationChallenge`** persists an **id-proofing snapshot** (`ProofingDateOfBirth`, `ProofingIdType`, `ProofingIdValue`) and **`DocvTokenIssuedAt`** so we can **re-run** `RunIdProofingAssessmentAsync` and obtain a **fresh DocV session** without making the user re-submit the form (when the snapshot exists).
- **`SubmitIdProofingCommandHandler`** saves that snapshot (and issuance time when a DocV session exists) on challenge creation.
- **`StartChallengeCommandHandler`**: if **Pending** or **Created** with stored token/URL and the token is **stale**, refreshes via Socure; clears **`SocureEventId`** on refresh; optional household enrichment for assessment; missing snapshot → conflict asking user to re-submit.
- **EF migration** `20260421235804_AddDocvTokenRefreshFieldsToDocVerificationChallenges`: adds nullable columns `DocvTokenIssuedAt`, `ProofingDateOfBirth`, `ProofingIdType`, `ProofingIdValue` on **`DocVerificationChallenges`**.
- **Repository:** removed **`DuplicateRecordException`** handling for the one-active-challenge unique index; concurrent duplicate inserts may surface as a generic DB error.

**Docs / config:** `appsettings.dc.example.json` documents and sets `DocvTransactionTokenTtlMinutes: 20` (code default applies if omitted).

#### 🔗 Links to related PRs
N/A

#### ✅ Completion tasks

- [x] Added relevant tests (`StartChallengeCommandHandlerTests`, `DocVerificationChallengeTests`, `SocureSettingsValidatorTests`, factory updates)
- [ ] Configuration changes:
  - [ ] If new environment variables are added, update in Tofu — **N/A** (appsetting only; optional)
  - [ ] If new environment secrets are added, update in Tofu and set in AWS Secret Manager — **N/A**
  - [x] If you're adding an appsetting, add it to the requisite example file, and update it in the AWS AppConfig — **example updated**; confirm AppConfig if non-default TTL is needed in deployed envs
  - [ ] If appsettings are changed, update in AWS AppConfig — **only if overriding default TTL**

**Deployment:** Apply migration `20260421235804_AddDocvTokenRefreshFieldsToDocVerificationChallenges` to all portal DBs (API startup migrations or explicit `dotnet ef database update`).

**Manual QA:** Id-proofing → doc-verify → `GET .../challenges/{id}/start` returns **200**; optionally force stale `DocvTokenIssuedAt` and confirm a second start refreshes the token (logs / new session).

[DC-321]: https://codeforamerica.atlassian.net/browse/DC-321?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


src/SEBT.Portal.Infrastructure/Repositories/DatabaseDocVerificationChallengeRepository.cs
```diff
@@ -114,16 +113,7 @@ await dbContext.DocVerificationChallenges
             var entity = MapToEntity(challenge);
             dbContext.DocVerificationChallenges.Add(entity);
 
-            try
```
necampanini, line 117: the catch block that translated the unique-index violation on `IX_DocVerificationChallenges_OneActivePerUser` into DuplicateRecordException was deleted, but the migration that creates that index is still around, and SubmitIdProofingCommandHandler.cs:237 still has `catch (DuplicateRecordException)` that expects the translation. 

in a concurrent insert race (double-click/retries/etc), we'd get a 500 instead of just re-querying the winner.

src/SEBT.Portal.Infrastructure/Migrations/PortalDbContextModelSnapshot.cs
```diff
@@ -91,6 +94,18 @@ protected override void BuildModel(ModelBuilder modelBuilder)
                         .HasMaxLength(255)
                         .HasColumnType("nvarchar(255)");
 
+                    b.Property<string>("ProofingDateOfBirth")
```
necampanini, line 97: Will this be handling raw PII? If so, should we consider encrypting? I know the SoCure call requires raw values but for example, `StartChallengeCommandHandler.cs:264` logs ex as is, and any DbUpdateException carrying column values in the message could leak.

necampanini (changes requested):
solid PR, I would just want verification on the unencrypted PII and `IX_DocVerificationChallenges_OneActivePerUser` constraint being removed from a try/catch but still existing in the DB

src/SEBT.Portal.UseCases/IdProofing/StartChallenge/StartChallengeCommandHandler.cs
```diff
@@ -159,4 +198,138 @@ public async Task<Result<StartChallengeResponse>> Handle(
...
+                assessment.Outcome, challenge.UserId);
+            return Result<StartChallengeResponse>.PreconditionFailed(
+                PreconditionFailedReason.Conflict,
+                "Unable to refresh document verification. Please re-submit your information.");
+        }
+
+        var session = assessment.DocvSession;
+        challenge.SocureReferenceId = session.ReferenceId;
+        challenge.EvalId = session.EvalId;
+        challenge.DocvTransactionToken = session.DocvTransactionToken;
```
necampanini, line 323: with a newly minted token, should `ExpiresAt` extend too? `Pending` challenges bypass the check already since a check earlier only looks at `Created`

src/SEBT.Portal.UseCases/IdProofing/SubmitIdProofing/SubmitIdProofingCommandHandler.cs
```diff
@@ -227,7 +228,11 @@ private async Task<Result<SubmitIdProofingResponse>> CreateChallengeAndRespond(
             DocvTransactionToken = assessment.DocvSession?.DocvTransactionToken,
             DocvUrl = assessment.DocvSession?.DocvUrl,
             SocureReferenceId = assessment.DocvSession?.ReferenceId,
-            EvalId = assessment.DocvSession?.EvalId
+            EvalId = assessment.DocvSession?.EvalId,
```
necampanini, line 231: I believe this is the only place the snapshot fields get populated, so it's worth asserting in a test in case down the line fields are dropped or renamed. if they were, StartChallenge would silently return `Conflict / "Please re-submit"`, and it'd be difficult to discern from a legitimate expired session. One extra Arg.Is<> assertion on the existing DocumentVerificationRequired test covers it

src/SEBT.Portal.Infrastructure/Repositories/DatabaseDocVerificationChallengeRepository.cs
noxmwalsh: Ack, I can't believe I missed that!  Thanks for pointing this out: this is addressed in a184c8f77a28fdf1d6db1ae435169b293624d11c

src/SEBT.Portal.Infrastructure/Migrations/PortalDbContextModelSnapshot.cs
noxmwalsh: Yeah, that `ex` object in the log shouldn't be displayed (this is addressed in 65cd74d5a5cef9653d0158a4544d0ac76bf57744).  I'm looking into a better solution for storing data in the DOB column right now.

noxmwalsh: I think we should handle encrypting the DOB column separately (we absolutely need to tackle though; we need to do the same for the User table columns before launch).  For now though, I can have it so we'll scrub out the DOB once the challenge is in a state where we don't need it anymore.

src/SEBT.Portal.UseCases/IdProofing/StartChallenge/StartChallengeCommandHandler.cs
noxmwalsh: Addressed in. 234e0d137dfe8c12fb60a982fc29d65479a7cfa2

src/SEBT.Portal.Infrastructure/Migrations/PortalDbContextModelSnapshot.cs
noxmwalsh: Addressed PII handling on the table with by scrubbing it out when we don't need it anymore in commit 0df514b43115a6562a442a83500d724a2ab3e1cc.  It's still ideal if we can encrypt the column, but we also should do that at the same time as we do for the User table.

src/SEBT.Portal.UseCases/IdProofing/SubmitIdProofing/SubmitIdProofingCommandHandler.cs
noxmwalsh: Addressed in 624b50a1d96a9fc71c29b80912d463ce7dab86bc.  Two tests were added

src/SEBT.Portal.Infrastructure/Migrations/PortalDbContextModelSnapshot.cs
noxmwalsh: New ticket for handling at-rest encryption https://codeforamerica.atlassian.net/browse/DC-324

necampanini (approved):

PR #205, opened 2026-04-22
DC-313: Enforce data annotation validation on startup for options classes

jamesmblair:
#### 🔗 Jira ticket
DC-313

#### ✍️ Description
`AddOptionsWithValidateOnStart<T>()` only controls *when* validation runs (at startup vs. on first access) — it does **not** register any validators. Without an explicit `.ValidateDataAnnotations()` call, the `[Required]`, `[MinLength]`, `[Range]`, and `[EmailAddress]` attributes on settings classes were decorative and never enforced.

**Impact:** The app could start with invalid configuration (e.g., empty `JwtSettings.SecretKey`) and pass health checks, then fail at runtime on first use. The crypto library (`SymmetricSecurityKey`) does reject empty/short keys at token-mint time, so tokens were never signed with an empty key — but the app shouldn't start at all with invalid config.

**Changes:**
- Added `.ValidateDataAnnotations()` to all 8 settings registrations that carry data annotations (`JwtSettings`, `IdentifierHasherSettings`, `EmailOtpSenderServiceSettings`, `OtpRateLimitSettings`, `EnrollmentCheckRateLimitSettings`, `WebhookRateLimitSettings`, `SocureSettings`, `SmartySettings`)
- Added unit tests proving the crypto library rejects empty/short JWT signing keys
- Added integration tests proving the app now refuses to start without a valid SecretKey
- Added JSONC comment to `appsettings.json` explaining why `SecretKey` is intentionally empty (overridden by Secrets Manager via Tofu)
- Configured VS Code workspace to treat `appsettings*.json` as JSONC so comments don't trigger lint warnings

#### 🔗 Links to related PRs
None

#### ✅ Completion tasks

- [x] Added relevant tests
- [ ] Configuration changes:
  - [x] If new environment variables are added, update in Tofu — N/A, no new env vars
  - [x] If new environment secrets are added, update in Tofu and set in AWS Secret Manager — N/A, no new secrets
  - [x] If you're adding an appsetting, add it to the requisite example file, and update it in the AWS AppConfig — N/A, no new appsettings
  - [x] If appsetttings are changed, update in AWS AppConfig — Only a comment was added, no value changes

### DIFF
--- a/sebt-portal.code-workspace
+++ b/sebt-portal.code-workspace
@@ -37,6 +37,7 @@
     "files.associations": {
       "*.scss": "scss",
       "*.json": "json",
+      "appsettings*.json": "jsonc",
     },
     "editor.defaultFormatter": "esbenp.prettier-vscode",
     "editor.formatOnSave": true,
--- a/src/SEBT.Portal.Api/appsettings.json
+++ b/src/SEBT.Portal.Api/appsettings.json
@@ -62,6 +62,7 @@
     "SecretKey": "OverrideInProductionUseEnvVarIDENTIFIERHASHER__SECRETKEY"
   },
   "JwtSettings": {
+    // Intentionally empty — overridden by Secrets Manager via Tofu in deployed environments
     "SecretKey": "",
     "Issuer": "SEBT.Portal.Api",
     "Audience": "SEBT.Portal.Web",
--- a/test/SEBT.Portal.Tests/Unit/Infrastructure/Services/JwtSecretKeyValidationTests.cs
+++ b/test/SEBT.Portal.Tests/Unit/Infrastructure/Services/JwtSecretKeyValidationTests.cs
@@ -32,7 +32,8 @@ public class JwtSecretKeyValidationTests
             new IdProofingValiditySettings { ValidityDays = 1826 },
             NullLogger<OidcVerificationClaimTranslator>.Instance);
 
-        return new JwtTokenService(jwtOptions, validityOptions, translator);
+        return new JwtTokenService(jwtOptions, validityOptions, translator,
+            NullLogger<JwtTokenService>.Instance);
     }
 
     [Fact]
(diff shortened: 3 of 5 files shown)


src/SEBT.Portal.Api/appsettings.json
```diff
@@ -62,6 +62,7 @@
     "SecretKey": "OverrideInProductionUseEnvVarIDENTIFIERHASHER__SECRETKEY"
   },
   "JwtSettings": {
+    // Intentionally empty — overridden by Secrets Manager via Tofu in deployed environments
```
noxmwalsh, line 65: Is this using an implementation of JSONC now?

sebt-portal.code-workspace
```diff
@@ -37,6 +37,7 @@
     "files.associations": {
       "*.scss": "scss",
       "*.json": "json",
+      "appsettings*.json": "jsonc",
```
noxmwalsh, line 40: Answered my own question from https://github.com/codeforamerica/sebt-self-service-portal/pull/205/changes#r3127381474 :)

test/SEBT.Portal.Tests/Unit/Infrastructure/Services/JwtSecretKeyValidationTests.cs
```diff
@@ -0,0 +1,69 @@
+using Microsoft.Extensions.Logging.Abstractions;
+using Microsoft.Extensions.Options;
+using NSubstitute;
+using SEBT.Portal.Core.AppSettings;
+using SEBT.Portal.Infrastructure.Services;
+
+namespace SEBT.Portal.Tests.Unit.Infrastructure.Services;
```
noxmwalsh, line 7: How is this different from what's done in `PortalWebApplicationFactory`?

noxmwalsh (approved):
Approved, although I do have one comment wondering about one of the test files

test/SEBT.Portal.Tests/Unit/Infrastructure/Services/JwtSecretKeyValidationTests.cs
jamesmblair: This is a unit test validating the behavior of JwtTokenService specifically. The tests for PortalWebAppliactionFactory are ensuring the settings are eagerly validated, resulting in an app-startup crash.

noxmwalsh: Ok, thanks for the clarification!

PR #207, opened 2026-04-23
DC-326: mock state address updates when UseMockHouseholdData is enabled

jamesmblair:
#### 🔗 Jira ticket
[DC-326](https://codeforamerica.atlassian.net/browse/DC-326)

#### ✍️ Description
When `UseMockHouseholdData` was enabled (e.g. DC dev environment), household **reads** correctly used the in-memory `MockHouseholdRepository`, but address **writes** still routed to the real state connector plugin (`DcAddressUpdateService`). This caused a 502 error ("The address update service is temporarily unavailable") because the DC backend database was unreachable.

**Root cause:** Asymmetry in the mock/real implementation selection — `UseMockHouseholdData` only gated `IHouseholdRepository`, not `StatesPlugins.Interfaces.IAddressUpdateService`.

**Fix:**
- Added `MockStateAddressUpdateService` that maps the plugin address to the core address and delegates to `MockHouseholdRepository.TryUpdateAddress`
- Added `TryUpdateAddress` method to `MockHouseholdRepository` that finds a household by identifier (email or phone) and updates both `AddressOnFile` and per-case `MailingAddress` in memory
- Conditionally registers the mock in `AddPlugins()` when `UseMockHouseholdData` is true, overriding the plugin-provided `IAddressUpdateService`
- Subsequent reads reflect the updated address in the UI

Also adds pnpm script preference guidance to CLAUDE.md.

#### 🔗 Links to related PRs
None

#### ✅ Completion tasks

- [x] Added relevant tests
- [x] Configuration changes:
  - [x] No new appsettings (uses existing `UseMockHouseholdData`)
  - [x] No appsettings changes

[DC-326]: https://codeforamerica.atlassian.net/browse/DC-326?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


CLAUDE.md
jamesmblair, line 1: Minor additions based on learnings from work on this PR.

noxmwalsh:
Yeah, I know this issue was going to come up; the mocks were not originally made with the thought that we'd need to persist and update them.  It's a bit tricky since the original purpose of them was to unblock frontend development while we were waiting for integration for DC/CO, but I can also see the need for having this for future states and to not have the app crumble on any write operations in state-agnostic mode.

I do think this is fine for now to unblock us, but we have to cautious on how far we want to go since I think there's more value in testing the full integration of DC/CO on Dev UAT and (if we get one) staging.

noxmwalsh (approved):
Approved with comment about personal concerns, but nothing that should block this from moving forward

PR #208, opened 2026-04-23
DC-214(feat): Cohort management for mixed-eligibility/applicant households

noxmwalsh:
#### 🔗 Jira ticket
[DC-214: Exclude mixed‑eligibility/applicant households from co‑loaded views](https://codeforamerica.atlassian.net/browse/DC-214)

#### ✍️ Description
Adds a derived-at-runtime cohort classification that formalizes which households
are excluded from co-loaded views, and wires it through to frontend analytics so
the excluded cohort can be measured distinctly.

**Backend**
- New `CoLoadedCohort` enum in Core (`NonCoLoaded` / `CoLoadedOnly` /
  `MixedOrApplicantExcluded`) attached to `HouseholdData`.
- `GetHouseholdDataQueryHandler` now classifies the household on its
  **pre-filter** state, then suppresses co-loaded cases only for the
  `MixedOrApplicantExcluded` cohort. This extends the prior mixed-eligibility
  filter to additionally cover applicants with co-loaded benefits (households
  whose only cases are co-loaded but who have pending applications or pending
  `ApplicationStatus` on a case).
- `HouseholdDataResponse` + mapper now expose `coLoadedCohort` on the wire.

**Frontend**
- `HouseholdDataSchema` adds `coLoadedCohort` (Zod enum w/ int→string preprocess,
  defaulting to `NonCoLoaded` for backward compatibility with older payloads).
- New `toAnalyticsCohort` helper maps the enum to the standardized snake_case
  analytics values.
- `DashboardContent` emits `co_loaded_cohort` as a user-scoped data-layer
  property on dashboard load (`non_co_loaded` / `co_loaded_only` /
  `mixed_or_applicant_excluded`), alongside the existing
  `household_linked_children` user property.

#### 🔗 Links to related PRs
There's no other PRs, but this does related to what was done from ADR-0012, so I recommend looking through that.

#### ✅ Completion tasks

- [x] Added relevant tests
  - 5 new backend handler tests in
  - 4 new frontend tests 
  - New schema tests in `schema.test.ts` for `CoLoadedCohortSchema` and
    `toAnalyticsCohort`
  - Updated existing fixtures/mock data in `testing/fixtures.ts` and six component tests
    updated for the new required field
  - ✅ Configuration-driven mechanism w/ documented update process
  - ✅ Backend suppresses co-loaded benefits for flagged households ->
    `GetHouseholdDataQueryHandler` filter now cohort-gated
  - ✅ Dashboard renders standard non-co-loaded experience, no UI hints -> no
    new copy; post-filter payload only
  - ✅ Analytics distinguish the three cohorts via standardized property ->
    `co_loaded_cohort` user-scoped data layer key
  - ✅ Tests verify all three household types in payloads and UI
- [x] Configuration changes:
  - [x] If new environment variables are added, update in Tofu *(N/A)*
  - [x] If new environment secrets are added, update in Tofu and set in AWS Secret Manager *(N/A)*
  - [x] If you're adding an appsetting, add it to the requisite example file, and update it in the AWS AppConfig *(N/A; no appsettings)*
  - [x] If appsetttings are changed, update in AWS AppConfig *(N/A; no appsettings changed)*


src/SEBT.Portal.UseCases/Household/GetHouseholdData/GetHouseholdDataQueryHandler.cs
```diff
@@ -116,7 +121,41 @@ public async Task<Result<HouseholdData>> Handle(GetHouseholdDataQuery query, Can
...
         return Result<HouseholdData>.Success(householdData);
     }
+
+    /// <summary>
+    /// Classifies the household based on its pre-filter case list and applications.
+    /// See <see cref="CoLoadedCohort"/> for the rule
+    /// The rule is intentionally derived at runtime so ops
+    /// changes to the cohort definition require only a code change, not a data migration.
+    /// </summary>
+    private static CoLoadedCohort ClassifyCoLoadedCohort(HouseholdData household)
```
spokenbird, line 142: AC # 1 asks for this to be configurable. I think we could add an AppConfig setting to turn the filter on and off?

src/SEBT.Portal.UseCases/Household/GetHouseholdData/GetHouseholdDataQueryHandler.cs
```diff
@@ -116,7 +121,41 @@ public async Task<Result<HouseholdData>> Handle(GetHouseholdDataQuery query, Can
...
+    {
+        var hasCoLoaded = household.SummerEbtCases.Any(c => c.IsCoLoaded);
+        if (!hasCoLoaded)
+        {
+            return CoLoadedCohort.NonCoLoaded;
+        }
+
+        var hasNonCoLoaded = household.SummerEbtCases.Any(c => !c.IsCoLoaded);
+        var hasApplications = household.Applications.Count > 0;
```
spokenbird: Question I'm uncertain about - is this too broad? For example the mock data has the co-loaded scenario where ApplicationStatus is approved but the scenario also has 2 co-loaded cases, as I understand it this would make it so they are classified as MixedOrApplicantExcluded?

My understanding of how that should work to be fair is not super firm, so feel free to push back if you think this is the desired behavior. 

If what I mention IS an issue, we may also want to add a test around it.

src/SEBT.Portal.UseCases/Household/GetHouseholdData/GetHouseholdDataQueryHandler.cs
```diff
@@ -73,13 +73,18 @@ public async Task<Result<HouseholdData>> Handle(GetHouseholdDataQuery query, Can
...
-        // their cases (they're all the user has), but per-case flags prevent actions.
-        // MVP intent confirmed by product: mixed households are not visually supported.
+        // Classify the household on the PRE-filter state so analytics can
+        // distinguish cohorts even after co-loaded cases are suppressed. Then
+        // apply the suppression for the excluded cohort.
+        householdData.CoLoadedCohort = ClassifyCoLoadedCohort(householdData);
+
         var nonCoLoaded = householdData.SummerEbtCases.Where(c => !c.IsCoLoaded).ToList();
-        if (nonCoLoaded.Count > 0)
+        if (householdData.CoLoadedCohort == CoLoadedCohort.MixedOrApplicantExcluded)
```
spokenbird: Before this change a household with only non-co-loaded cases would have it's BenefitIssuanceType set to SummerEbt. 

After this change, I believe it would be set to whatever we get back from the connector which may be fine? But I wanted to call it out just in case. 

We could also potentially add an assert to Handle_ClassifiesCohort_AsNonCoLoaded_WhenNoCasesAreCoLoaded

src/SEBT.Portal.UseCases/Household/GetHouseholdData/GetHouseholdDataQueryHandler.cs
```diff
@@ -116,7 +121,41 @@ public async Task<Result<HouseholdData>> Handle(GetHouseholdDataQuery query, Can
...
+        logger.LogDebug(
+            "Household data retrieved successfully for identifier type {Type}, cohort {Cohort}",
+            identifier.Type,
+            householdData.CoLoadedCohort);
         return Result<HouseholdData>.Success(householdData);
     }
+
+    /// <summary>
+    /// Classifies the household based on its pre-filter case list and applications.
+    /// See <see cref="CoLoadedCohort"/> for the rule
```
spokenbird:
```suggestion
    /// See <see cref="CoLoadedCohort"/> for the rule.
```

src/SEBT.Portal.Web/src/features/household/api/schema.ts
```diff
@@ -228,7 +264,11 @@ export const HouseholdDataSchema = z.object({
   addressOnFile: AddressSchema.nullable().optional(),
   userProfile: UserProfileSchema.nullable().optional(),
   benefitIssuanceType: IssuanceTypeSchema.nullable().optional(),
-  allowedActions: AllowedActionsSchema.nullable().optional()
+  allowedActions: AllowedActionsSchema.nullable().optional(),
+  // Defaults to NonCoLoaded for backward compatibility with older/cached payloads
+  // that may not carry the field yet; that's the safest assumption because it
+  // means the excluded-cohort analytics dimension is not falsely asserted.
+  coLoadedCohort: CoLoadedCohortSchema.optional().default('NonCoLoaded')
```
spokenbird: Suggestion (non-blocking):

The default here silently maps a missing field to a real cohort value (NonCoLoaded), collapsing "genuinely non-co-loaded user" and "field didn't arrive" into the same analytics signal, meaning the excluded cohort would be silently under counted if the field ever stops arriving. 

This could be a non-issue since we always expect the field, but for defense in depth it would be good to know if the co-loaded cohort analytics value ever stops reflecting reality should someone make a breaking change down the road.

I'm thinking we could either explicitly error if the field is missing, so something like:
`coLoadedCohort: CoLoadedCohortSchema` which would error on the field missing and we'd know right away. 

Or alternatively, default to something like Unknown so it shows up as a distinct value if it ever happens?

spokenbird (changes requested):
Over all looks solid, two main asks, configurability for AC#1 and clarification on whether hasApplications is too broad. Smaller notes / nits inline.

src/SEBT.Portal.UseCases/Household/GetHouseholdData/GetHouseholdDataQueryHandler.cs
noxmwalsh: Oh right, I forgot to add this in.  I'll do that

scripts/git/checkout-branch-all-repos.sh
```diff
@@ -0,0 +1,55 @@
+#!/usr/bin/env bash
```
noxmwalsh, line 1: This is an unrelated helper script I decided to add to make it easy to just create/checkout all branches at once for every repo.

src/SEBT.Portal.UseCases/Household/GetHouseholdData/GetHouseholdDataQueryHandler.cs
noxmwalsh: Addressed in 8c1fc52dad9f9014ab581cb009ae21faaff741f8

src/SEBT.Portal.UseCases/Household/GetHouseholdData/GetHouseholdDataQueryHandler.cs
noxmwalsh: I think you're right about this.  I can update this case to cover that;  This is now addressed in d313fc6a31e4deccf4ea24c964d00485ebb00983

src/SEBT.Portal.UseCases/Household/GetHouseholdData/GetHouseholdDataQueryHandler.cs
noxmwalsh: We can be sure about this by updating the tests.  This is done in d313fc6a31e4deccf4ea24c964d00485ebb00983

src/SEBT.Portal.Web/src/features/household/api/schema.ts
noxmwalsh: I went with it by default setting it to Unknown.  This change is in 58428ff5684e1d1b9ff0dbd36d5a27dbb830d7db

spokenbird (approved):

PR #213, opened 2026-04-24
DC-212(feat): co-loaded no children / no benefits terminal path

adbergen:
#### 🔗 Jira ticket
https://codeforamerica.atlassian.net/browse/DC-212

#### ✍️ Description
Wires the two terminal-path screens for co-loaded DC users:

- **NOT_FOUND** — when the backend cannot match the submitted SNAP/TANF + DOB, the off-boarding page now renders the shared "We're sorry, we aren't able to show your DC SUN Bucks information" copy with buttons styled per Figma (outline "Back" on the left, filled "Contact us" on the right, "Apply now" below). Co-loaded users never reach Socure DocV — the back action returns to the co-loaded ID entry form.
- **NO_CHILDREN** — a co-loaded user whose ID matches but whose household has zero children and zero applications lands on the dashboard's existing warning alert. A new `co-loaded-no-children` seed scenario (DC-only, SNAP ID `SNAP-CO-NC-001`) makes this path testable in local dev.
- **Analytics** — `idv_primary_result` now includes an `idv_primary_reason` page-data field (`not_found` for co-loaded mismatches, `socure_fail` otherwise). `household_result` includes `household_reason: 'no_children'` when a co-loaded user lands on the empty dashboard. Both fire exactly once per attempt.

Stacked on DC-88 — needs DC-88 merged first. Base branch for this PR is `feature/DC-88-co-loaded-identity-proofing`; once DC-88 lands on main, rebase the base to `main`.

Note: the Figma shows the short label "Enter an ID number" on the back button, but the DC `offBoarding.action` CSV `Current` column still has "Go back and enter an ID number" (only the `SOURCE English` column has the short form). Content-team update to the Google Sheet needed to align the `Current` column with the Figma.

#### 🔗 Links to related PRs
- DC-88: https://github.com/codeforamerica/sebt-self-service-portal/pull/200

#### ✅ Completion tasks

- [x] Added relevant tests
- [x] Configuration changes:
  - [ ] If new environment variables are added, update in Tofu — none added
  - [ ] If new environment secrets are added, update in Tofu and set in AWS Secret Manager — none added
  - [ ] If you're adding an appsetting, add it to the requisite example file, and update it in the AWS AppConfig — none added
  - [ ] If appsetttings are changed, update in AWS AppConfig — none changed


adbergen:
# QA Summary — DC-212 Co-Loaded "No Children / No Benefits" Terminal Path

## What Changed

Co-loaded DC users (those authenticating with DOB + SNAP/TANF ID instead of Socure) now see the correct shared error screens when the backend cannot identify them or when their household has no enrolled children. They are blocked from entering the portal and can only retry via the co-loaded ID entry form, never the Socure flow. The off-boarding copy, button styles, and button layout have been updated to match the Figma (left "Enter an ID number" outline button, right "Contact us" filled button, with an "Apply now" section below). A new seed user has been added so the "matched but no children" dashboard empty state is reachable in local dev.

## Affected Areas

- Login → Enter an ID number screen, plus the off-boarding screen shown after a failed submission
- Dashboard landing page for authenticated users with no enrolled children and no applications
- Analytics data layer (failure reason tagging)

## Test Cases

### Happy Path

1. **Co-loaded user sees the "cannot identify you" screen on ID mismatch**
   - Sign in as `sebt.dc+co-loaded-pending-id-proofing@codeforamerica.org`
   - On the "Enter an ID number" screen, enter any DOB and any SNAP/TANF ID that will NOT match backend records (e.g., `WRONG-999`)
   - Click Continue
   - Expected: screen titled "We're sorry, we aren't able to show your DC SUN Bucks information"
   - Expected: body text "You can go back to enter an ID number, or contact us if you need more help."
   - Expected: two buttons visible — "Enter an ID number" (outline) on the left, "Contact us" (filled yellow) on the right
   - Expected: below, an "Apply now" section with body text ending "Tap 'Apply now' and enter your child's information."
   - Expected: URL-hacking to `/dashboard` does not reach the dashboard

2. **"Enter an ID number" button returns to the ID entry form (NOT Socure)**
   - From the screen in test 1, click "Enter an ID number"
   - Expected: returned to the "Enter an ID number" form
   - Expected: DOB and ID fields are empty and re-submittable

3. **"Apply now" button navigates to the apply flow**
   - From the screen in test 1, click "Apply now"
   - Expected: navigates to the apply page (same destination as the dashboard's empty-state "Apply for DC SUN Bucks now" link)

4. **Co-loaded user with no children lands on the dashboard empty state**
   - Sign in as `sebt.dc+co-loaded-no-children@codeforamerica.org` (new seed added with this ticket)
   - On the "Enter an ID number" screen, enter any valid DOB and SNAP Account ID = `SNAP-CO-NC-001`
   - Click Continue
   - Expected: backend matches and redirects to the dashboard
   - Expected: only the user profile card (initials + name + Logout link) and a yellow warning alert "Currently, you have no children enrolled in DC SUN Bucks or DC SUN Bucks applications" are visible
   - Expected: alert includes body text "Come back to this dashboard in 1–2 business days…" and a link "Apply for DC SUN Bucks now"
   - Expected: no household summary card, no enrolled children list, no EBT Edge section, no applications section, and no action buttons appear

### Edge Cases

1. **Co-loaded user on Spanish or Amharic locale sees the translated error copy**
   - Repeat happy path test 1 with the language toggle set to Español
   - Expected: title and body appear in Spanish; Back and Contact us buttons remain functional
   - Repeat with Amharic if available in the environment

2. **Button layout on narrow and wide viewports**
   - View the off-boarding screen on a narrow mobile viewport (< 480px)
   - Expected: buttons stack vertically, each full-width, text fits on one line
   - Resize to tablet width
   - Expected: buttons sit side-by-side

3. **Event fires exactly once per submit attempt**
   - Submit an incorrect ID and DOB as a co-loaded user twice in a row (using the Back button between)
   - Expected: each attempt produces exactly one redirect to off-boarding
   - If the analytics data-layer inspector is available, confirm a single `idv_primary_result` event per submit, with reason `not_found` on the co-loaded path

4. **canApply=false hides the Apply section**
   - If the test environment can force the backend to return `canApply: false` for a co-loaded user, submit a failing attempt
   - Expected: title, body, and the two top buttons still appear
   - Expected: the "Apply now" section is hidden

### Regression Checks

1. **Non-co-loaded Socure failure shows the original DocV-flavored off-boarding copy**
   - Sign in as `sebt.dc+non-co-loaded@codeforamerica.org`
   - Trigger a Socure failure
   - Expected: title "We want to keep your account safe" (not the co-loaded "We're sorry…" copy)
   - Expected: behavior is unchanged from before this ticket

2. **Non-co-loaded user with empty dashboard still sees the standard empty-state alert**
   - Sign in as a non-co-loaded user who has zero enrolled children and zero applications
   - Expected: the same yellow warning alert and "Apply for DC SUN Bucks now" link appear as before

3. **Document verification flow is unaffected**
   - Trigger a Socure path that requires document verification
   - Expected: redirected to the document upload screen as before

4. **Dashboard with cases renders normally**
   - Sign in as a user (co-loaded or not) who has enrolled children
   - Expected: full dashboard renders — household summary, children list, EBT Edge section, applications, action buttons

5. **Co-loaded happy path to a populated dashboard still works**
   - Sign in as `sebt.dc+co-loaded@codeforamerica.org` (already ID-verified)
   - Expected: lands on the full dashboard with enrolled children, no empty-state alert

## Environment Notes

- Depends on DC-88 (co-loaded identity proofing) being deployed to the same environment; this ticket cannot be tested without it
- Requires a DC-configured environment; this flow does not exist for CO
- Local dev: restart the API after pulling this branch so the new `co-loaded-no-children` seed user is created. If it doesn't appear in Mailpit, run the ClearSeededData console app and restart to re-seed from scratch
- Test users needed:
  - `sebt.dc+co-loaded-pending-id-proofing@codeforamerica.org` — for the "cannot identify" path; enter any mismatched SNAP ID
  - `sebt.dc+co-loaded-no-children@codeforamerica.org` — for the "no children" dashboard path; enter SNAP Account ID `SNAP-CO-NC-001`
  - `sebt.dc+co-loaded@codeforamerica.org` — for the populated dashboard regression
  - `sebt.dc+non-co-loaded@codeforamerica.org` — for the Socure regression path

src/SEBT.Portal.Infrastructure.Seeding/Services/DatabaseSeeder.cs
```diff
@@ -345,6 +364,21 @@ public void SeedTestUsers(bool useMockHouseholdData)
...
                     }
+                    else if (normalizedEmail == coLoadedNoChildrenEmail)
+                    {
+                        user = UserFactory.CreateCoLoadedUser(u =>
+                        {
+                            u.Email = normalizedEmail;
+                            u.IdProofingStatus = IdProofingStatus.Completed;
+                            u.IalLevel = scenario.IalLevel;
+                            u.IdProofingCompletedAt = now.AddDays(DaysSinceIdProofingCompleted);
+                            u.CoLoadedLastUpdated = now.AddDays(DaysSinceCoLoadedUpdate);
```
jamesmblair, line 375: I'm unclear on what this attribute is about and why this specific scenario is needed. I believe the API is computing at a *case* level whether the case is co-loaded and we should just be reading that for UI flow decisions.

jamesmblair: Or are we writing this timestamp when a user submits their IC/DOB and we do a check with those values against the backend? Just trying to confirm my understanding of what exactly this means within the domain model.

src/SEBT.Portal.Web/src/app/(public)/login/id-proofing/off-boarding/page.tsx
```diff
@@ -1,40 +1,66 @@
...
-}
+export default function OffBoardingPage() {
+  const searchParams = useSearchParams()
+  const canApply = searchParams.get('canApply') !== 'false'
 
-export default async function OffBoardingPage({ searchParams }: OffBoardingPageProps) {
-  const params = await searchParams
-  const canApply = params.canApply !== 'false'
+  const { session } = useAuth()
+  const isCoLoaded = session?.isCoLoaded === true
```
jamesmblair, line 15: Related to my question about backend data/mocks. I thought this was being evaluated in the model at a *case* level. Is this an aggregation or something?

adbergen: @jamesmblair it's the JWT is_co_loaded claim, which comes from User.IsCoLoaded on the portal's User table. We need it here because the off-boarding page runs during login, before household data is fetched, so case-level info isn't available yet.

src/SEBT.Portal.Infrastructure.Seeding/Services/DatabaseSeeder.cs
adbergen: @jamesmblair User.IsCoLoaded and CoLoadedLastUpdated are user-level fields populated from external system separate from SummerEbtCase.IsCoLoaded, which is computed per-case from IssuanceType. Using the user-level one in the auth flow (login, id-proofing) where we don't have household data yet. Not written on IC/DOB submission. The seeder is just matching the production User-row shape.

jamesmblair: Right. Reviewing the full codebase though - I'm not seeing anywhere these values get set besides fake/mocked data. I think there's an inversion/misunderstanding possibly happening. We don't know if a user has co-loaded cases until we try to match the identifiers they provide in the UI with the backend system. I believe *all* users go through the UI screen where we ask for these identifiers, so we shouldn't need to know at this point whether they have co-loaded cases.

Happy to have a synchronous pairing/discussion if it would help!

src/SEBT.Portal.Web/src/app/(public)/login/id-proofing/off-boarding/page.tsx
adbergen: You're right — I checked and `User.IsCoLoaded` is initialized to `false` at user creation and never updated anywhere in production. It only gets set in mocks/seeders/tests, so this branch never fires in prod. Same for the co-loaded SNAP/TANF + DOB match in `SubmitIdProofingCommandHandler` and the `no_children` analytics tag in this PR.

Agree we should determine co-loaded status from the identifier-match flow. Going to wire the user-level flag to actually get set on successful match (drop the precondition gate and persist `IsCoLoaded=true` when the match returns), so the downstream UI logic becomes meaningful. Will push and re-request review. Happy to pair if anything's unclear.

src/SEBT.Portal.Infrastructure.Seeding/Services/DatabaseSeeder.cs
adbergen: You're right haha these are user-level fields that are never written in production. `User.IsCoLoaded` is initialized to `false` at user creation and nothing updates it anywhere; only mocks/seeders/tests set it. The co-loaded SNAP/TANF + DOB match path in `SubmitIdProofingCommandHandler` is also gated on this flag, so that path is dead in prod too.

Going to wire it up properly: drop the precondition gate so the match attempt happens for any user submitting a SNAP/TANF id, and set `User.IsCoLoaded=true` (with `CoLoadedLastUpdated`) on a successful match. The user-level flag becomes derived state from the match rather than a presumed-populated cache. Will push and re-request review.

src/SEBT.Portal.Web/src/app/(public)/login/id-proofing/off-boarding/page.tsx
adbergen: @jamesmblair  [ba1fd64](https://github.com/codeforamerica/sebt-self-service-portal/pull/213/commits/ba1fd64856b5ea94dbc70f40e498053b81e59ab6)

jamesmblair (approved):

PR #215, opened 2026-04-24
Bugfix DC-296 - Socure data validation/IAL elevation

necampanini:
# PR Review Priority Guide — DC-296

- **Branch:** `bugfix/DC-296-socure-data-validation` -> `main`
- **State:** pre-PR (no #)
- **Risk Level:** Medium-High
- **Files:** 30 | **Lines:** +2,207 / -127
- **Generated:** 2026-04-24

Catches Socure outbound payload shape errors (10-digit SSN/ITIN, non-E.164 phone, oversized names, reused evaluation ids) before they 400 the workflow, plus routes DocV webhook outcomes off the top-level workflow decision and refreshes JWT claims from DB post-DocV. Reviewer attention belongs on the claim-refresh logic, the webhook decision-routing shift, and the boundary validation guards.

---

#### 🔗 Jira ticket
- [DC-296](https://codeforamerica.atlassian.net/browse/DC-296)

#### ✍️ Description
Hardens the data validation on the outbound path to Socure so shape errors never reach the workflow. Frontend Zod schema enforces 9-digit SSN/ITIN and calendar-valid DOB. Backend adds the same SSN/ITIN guard in `SubmitIdProofingCommandHandler`, normalizes phones to E.164 and truncates names to 240 chars at the `HttpSocureClient` boundary, and sends a unique `id` per evaluation. DocV webhook processing routes on the top-level workflow decision (ACCEPT/REJECT/RESUBMIT/REVIEW) rather than the document-enrichment decision. Session JWT refresh re-pulls `ial` and `id_proofing_status` from the DB when the DB reflects a post-DocV elevation, using a rank comparison to block any accidental downgrade. Also cleans up offboarding copy for decline-consent and normalizes `SocureExcludedIdentifierTypes`.

#### 🔗 Links to related PRs
- Connector changes: none (portal-only)
- Follow-ups tracked: DC-138 resubmit handling (deferred), offboarding copy distinctness (tracked separately), Medicaid ID shape (open product question)

#### ✅ Completion tasks
- [x] Added relevant tests
- [x] Meets acceptance criteria in ticket
- [ ] Configuration changes:
  - [ ] If new environment variables are added, update in Tofu
  - [ ] If new environment secrets are added, update in Tofu and set in AWS Secret Manager
  - [ ] If you're adding an appsetting, add it to the requisite example file, and update it in the AWS AppConfig
  - [ ] If appsetttings are changed, update in AWS AppConfig

[...]

PR #218, opened 2026-04-27
DC-296 - Document Verification in new tab

necampanini:
#### 🔗 Jira ticket
[DC-296](https://codeforamerica.atlassian.net/browse/DC-296) 

#### ✍️ Description
Three follow-ups to PR #215, all on the DC ID-proofing → DocV path:

- **DocV capture: drop Socure WebSDK, hand off to hosted URL.** `handleContinue` now opens `about:blank` synchronously (preserving user-gesture activation), awaits the start-challenge JIT token, then redirects the new tab to `https://verify.socure.com/#/dv/<token>`. Same-tab fallback if the popup is blocked. The hosted URL is device-aware: desktop renders the QR for phone hand-off, mobile lands directly in camera capture, so the experience is consistent across desktop and mobile. Adapter/capture sub-state code is gone; orphaned adapter files left on disk for revert safety, future knip cleanup.
- **JWT refresh email lookup.** `GenerateForSessionRefresh` now resolves email through the `ClaimsPrincipal.GetUserEmail()` extension instead of a single dictionary key, so either the long URI claim form (`http://schemas.xmlsoap.org/...`) or short `email` form works. Addresses Copilot Finding #1 on PR #215.
- **DOB form error UX.** Composite-date schema failures (impossible calendar date like Feb 31, future date, age >120 years) previously attached the error message to the year input — flagging a perfectly valid year as invalid. Now they render at the fieldset level under the "Date of birth" legend, with `usa-form-group--error` on the fieldset and no `aria-invalid` on any single input.

#### 🔗 Links to related PRs
- Builds on #215 (merged) — DC-296 Socure data validation / IAL elevation

#### ✅ Completion tasks
- [x] Added relevant tests
- [x] Configuration changes:
  - [x] No new appsettings
  - [x] No appsettings changed

[DC-296]: https://codeforamerica.atlassian.net/browse/DC-296?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


src/SEBT.Portal.Web/src/features/auth/components/doc-verify/DocVerifyPage.tsx
```diff
@@ -87,61 +69,63 @@ export function DocVerifyPage({ contactLink, sdkKey }: DocVerifyPageProps) {
...
 
+    // Seed challengeId from URL/sessionStorage after mount because sessionStorage
+    // is not available during SSR — we can't lazy-init this via useState.
+    // eslint-disable-next-line react-hooks/set-state-in-effect
     setChallengeId(ctx.challengeId)
     // Persist to sessionStorage for mobile tab recovery
     sessionStorage.setItem(SK_CHALLENGE_ID, ctx.challengeId)
 
-    // If the user was in capture (e.g., mobile tab recovery), skip to pending (D6)
+    // Either persisted sub-state means the user already handed off to Socure
```
jamesmblair, line 79: Not something that would block this PR, because I don't think you've changed the logic for it, but is there a possibility where a document challenge times out and we *don't* want to automatically resume the pending state? Does the webhook handler account for that?

test/SEBT.Portal.Tests/Unit/Infrastructure/Services/SessionRefreshTokenServiceTests.cs
```diff
@@ -284,4 +284,25 @@ public void EmailFallsBackToEmpty_WhenNoEmailClaim()
         var jwt = ReadJwt(token);
         Assert.Equal("", jwt.Claims.First(c => c.Type == ClaimTypes.Email).Value);
     }
+
+    // BuildAndSignToken writes the email claim under ClaimTypes.Email and JWT bearer
+    // runs with MapInboundClaims=false, so the refresh principal carries the long URI
+    // form. GenerateForSessionRefresh must resolve the email from either form or the
+    // refreshed JWT will silently blank the user's email.
+    [Fact]
+    public void PreservesEmail_WhenPrincipalEmailUsesLongFormClaimType()
```
jamesmblair, line 293: We should really change this to issue our JWT with *only* the short form (`email`) as it is the standard form. Not something to address in this PR though.

jamesmblair (approved):
Approved with non-blocking questions/comments.

src/SEBT.Portal.Web/src/features/auth/components/doc-verify/DocVerifyPage.tsx
necampanini: if the document challenge times out, we get an `Expired` response from the polling endpoint, even if we start out in the UI as pending

PR #227, opened 2026-04-28
Chore/iis release package

jhernandez-cfa:
#### ✅ Completion tasks

- [ ] Added relevant tests
- [ ] Configuration changes:


scripts/ci/package-frontend.sh
```diff
@@ -0,0 +1,178 @@
...
+      shift
+      ;;
+  esac
+done
+
+log_info()    { echo -e "${BLUE}ℹ️  $1${NC}"; }
+log_success() { echo -e "${GREEN}✅ $1${NC}"; }
+log_warning() { echo -e "${YELLOW}⚠️  $1${NC}"; }
+log_error()   { echo -e "${RED}❌ $1${NC}"; }
+
```
jhernandez-cfa, line 62:
```suggestion

remove_dangling_symlinks() {
  local dangling_links_file
  dangling_links_file=$(mktemp)

  find "$STANDALONE_DIR" -type l ! -exec test -e {} \; -print > "$dangling_links_file"

  if [ -s "$dangling_links_file" ]; then
    log_warning "Removing dangling symlinks from standalone output before packaging..."
    while IFS= read -r dangling_link; do
      [ -n "$dangling_link" ] || continue
      log_warning "  $(basename "$dangling_link")"
      rm "$dangling_link"
    done < "$dangling_links_file"
  fi

  rm -f "$dangling_links_file"
}

```

scripts/ci/package-frontend.sh
```diff
@@ -0,0 +1,178 @@
...
+}
+
+package_frontend() {
+  STAGING_DIR=$(mktemp -d)
+  local WEB_DIR="$STAGING_DIR/sebt-web"
+
+  trap 'rm -rf "$STAGING_DIR"' EXIT
+
+  log_info "Staging portable build..."
+
```
jhernandez-cfa, line 105:
```suggestion

  remove_dangling_symlinks

```

.github/workflows/release-iis-dc.yaml
```diff
@@ -0,0 +1,187 @@
...
+        with:
+          fetch-depth: 0
+
+      - name: Resolve version
+        id: ver
+        run: |
+          if [[ "${GITHUB_REF}" == refs/tags/release/dc-v* ]]; then
+            VERSION="${GITHUB_REF#refs/tags/release/dc-v}"
+          elif [ -n "${{ github.event.inputs.version }}" ]; then
+            VERSION="${{ github.event.inputs.version }}"
```
jamesmblair: **Treat `inputs.version` as untrusted input**

`${{ github.event.inputs.version }}` is interpolated into the `run:` block at YAML expansion time, before the shell parses the script. GitHub's hardening guidance recommends not using context expressions inline in shell commands for inputs that aren't fully trusted, and instead passing them through environment variables: <https://docs.github.com/en/actions/security-guides/security-hardening-for-github-actions#using-an-intermediate-environment-variable>.

Suggested shape:

```yaml
- name: Resolve version
  id: ver
  env:
    INPUT_VERSION: ${{ github.event.inputs.version }}
  run: |
    if [[ "${GITHUB_REF}" == refs/tags/release/dc-v* ]]; then
      VERSION="${GITHUB_REF#refs/tags/release/dc-v}"
    elif [ -n "$INPUT_VERSION" ]; then
      VERSION="$INPUT_VERSION"
    else
      VERSION="dev-$(git rev-parse --short HEAD)"
    fi
    if [[ ! "$VERSION" =~ ^[A-Za-z0-9._-]+$ ]]; then
      echo "::error::Invalid version: $VERSION"
      exit 1
    fi
    echo "version=$VERSION" >> "$GITHUB_OUTPUT"
```

Same pattern for the other `run:` blocks that interpolate `${{ steps.ver.outputs.version }}` directly.

.github/workflows/release-iis-dc.yaml
```diff
@@ -0,0 +1,187 @@
...
+            VERSION="dev-$(git rev-parse --short HEAD)"
+          fi
+          echo "version=$VERSION" >> "$GITHUB_OUTPUT"
+          echo "git_sha=$(git rev-parse HEAD)" >> "$GITHUB_OUTPUT"
+          echo "Resolved version: $VERSION"
+
+      - name: Determine state-connector ref
+        id: state-connector-ref
+        run: |
+          BRANCH="${{ github.event_name == 'pull_request' && github.head_ref || github.ref_name }}"
```
jamesmblair: **Same env-var pattern for `github.head_ref`**

Branch names from PR head refs fall under the same untrusted-input category as `inputs.*` per GitHub's hardening guidance. Recommend the same env-var marshalling here so the value reaches the shell as a normal variable rather than a YAML-time substitution.

Behavior-preserving change — same URL, same two-branch fallback:

```yaml
- name: Determine state-connector ref
  id: state-connector-ref
  env:
    HEAD_REF: ${{ github.head_ref }}
    REF_NAME: ${{ github.ref_name }}
    EVENT_NAME: ${{ github.event_name }}
    CONNECTOR_PAT: ${{ secrets.SEBT_CONNECTOR_REPOSITORY_PAT }}
    REPO_OWNER: ${{ github.repository_owner }}
  run: |
    if [ "$EVENT_NAME" = "pull_request" ]; then
      BRANCH="$HEAD_REF"
    else
      BRANCH="$REF_NAME"
    fi
    FALLBACK="main"
    REPO_URL="https://x-access-token:${CONNECTOR_PAT}@github.com/${REPO_OWNER}/sebt-self-service-portal-state-connector.git"
    if git ls-remote --exit-code --heads "$REPO_URL" "refs/heads/${BRANCH}" 1>/dev/null 2>&1; then
      echo "ref=${BRANCH}" >> "$GITHUB_OUTPUT"
    else
      echo "ref=${FALLBACK}" >> "$GITHUB_OUTPUT"
    fi
```

Same shape for the dc-connector-ref step.

scripts/ci/bundle-iis-package.sh
```diff
@@ -0,0 +1,126 @@
...
+#       --web-zip <path> \
+#       --dacpac <path> \
+#       --changelog <path>          # CHANGELOG-DACPAC.md
+#       --deploy-report-xml <path>
+#       --deploy-report-html <path>
+#       --version <ver> \
+#       --git-sha <sha> \
+#       --output <zip>
+
+set -e
```
jamesmblair: **Add `set -o pipefail` to every script in `scripts/ci/`**

All seven scripts use `set -e; set -u` but omit `pipefail`. Default bash pipeline semantics return only the last command's exit code, so failures in `gh release list | grep | head`, `du | cut`, `find | while`, etc. are silently dropped.

One-line change per file: `set -euo pipefail` at the top. Worth re-running each script after the change since some `... || true` patterns may start protecting against real failures rather than just no-match cases.

Files: `bundle-iis-package.sh:18-19`, `extract-dacpac.sh:21-22`, `generate-dacpac-report.sh:24-25`, `package-frontend.sh:21-22`, `publish-api.sh:22-23`, plus the test scripts under `tests/` and the additions in `build-backend.sh`.

.github/workflows/release-iis-dc.yaml
```diff
@@ -0,0 +1,187 @@
+# Release IIS Package — DC
+# Builds an IIS deployment bundle for DC as an additional delivery option
+# alongside the existing Docker/ECR process: API + web + dacpac + README +
+# schema-change report.
+#
+# Triggers:
+#   workflow_dispatch (manual button, optional version input)
+#   push of tags matching release/dc-v* (creates a GitHub Release with the bundle attached)
```
jamesmblair: **Workflow header and the `Create GitHub Release` step disagree with the `on:` triggers**

Header comment says tag pushes matching `release/dc-v*` create a GitHub Release, but the tag trigger was removed in 1a075d11 and is no longer in `on:`. The `Create GitHub Release` step further down has an `if:` condition that can never fire as a result.

Plan: releases are created manually after `workflow_dispatch` for now. Suggest:

1. Update the header comment to reflect that.
2. Comment out (don't delete) the tag-related bits — the trigger and the `Create GitHub Release` step — so they're easy to revive later.

```yaml
# Triggers:
#   - workflow_dispatch (manual; produces a build artifact for download)
#   - pull_request (validation only — no artifact published)
#
# Releases are created manually from the workflow_dispatch artifact.
# The tag-push trigger and Create GitHub Release step are commented out
# below; revive them when we're ready to automate releases.
```

If the tag trigger comes back, it'll need the env-var fix from the version-input thread before it's safe to use.

jamesmblair (commented):
Let me know if you'd like to discuss any of the specific comments!

scripts/ci/templates/README.iis.md.tmpl
```diff
@@ -0,0 +1,121 @@
...
+5. Verify the API and web endpoints, then keep the prior deployment directory available for rollback.
+
+## Prerequisites on the IIS server
+
+Install once per IIS server. None of these runtime prerequisites are bundled in this zip.
+
+- **Windows Server** 2019+ with IIS enabled.
+- **ASP.NET Core 10 Hosting Bundle.** Download from <https://dotnet.microsoft.com/download/dotnet/10.0>. After install, run `iisreset` once.
+- **Node.js 24 LTS (x64)** — the MSI installer from <https://nodejs.org>. Confirm with `node -v`.
+- **HttpPlatformHandler module for IIS** — install via Web Platform Installer or download from Microsoft. Required for the web tier.
```
jamesmblair: Quick question on the WebPI reference here — what's our current understanding of how DC is using Web Platform Installer? Microsoft retired it on Dec 31 2022 ([IIS team announcement](https://blogs.iis.net/iisteam/web-platform-installer-end-of-support-feed)), so I want to make sure we know whether DC has it pre-installed from earlier work, has a known-good alternative path, or might hit a wall following this line as written. If you've already confirmed it works on their box, no change needed — just want to capture that context somewhere (here or in the PR description).

jhernandez-cfa: aligned with ms recomendation

jamesmblair (approved):

PR #240, opened 2026-04-30
DC-337: Loading interstitial for CO

adbergen:
#### 🔗 Jira ticket
https://codeforamerica.atlassian.net/browse/DC-337

#### ✍️ Description
Adds a "Please wait…" loading interstitial across four CO surfaces that previously left the user on a blank or unresponsive page during in-flight requests:

- **CO Enrollment Checker** — review page submit replaces the form with the interstitial until the eligibility response returns
- **CO Portal sign-in callback** — replaces the placeholder text during the OIDC code/token exchange
- **CO Portal AuthGuard** — replaces the previous blank `return null` while the auth-status fetch is in flight (covers refresh / direct URL navigation while signed in)
- **CO Portal Dashboard initial load** — replaces the existing `DashboardSkeleton` while household data is fetching

The interstitial is a new shared component in `@sebt/design-system` (`LoadingInterstitial`). Strings come from the existing CO `step-upProcessing` namespace (S10 in the CSV) — a one-line change to the enrollment checker's locale generator pulls those keys into its bundle. DC behavior is unchanged on every surface (DC keeps its `DashboardSkeleton` and the existing public-callback rendering).

#### 🔗 Links to related PRs
N/A

#### ✅ Completion tasks

- [x] Added relevant tests
- [ ] Configuration changes:


adbergen:
QA Summary — DC-337 Loading Interstitial (CO)

What Changed
Adds a "Please wait..." loading screen across four CO surfaces that previously left users on a blank or unresponsive page during in-flight requests: the Enrollment Checker submit step, the post-sign-in OIDC callback, the protected-route auth check, and the dashboard initial data fetch. The interstitial reads "Please wait..." with an info callout: "Do not exit the page. Checking to see if we have enough information." No backend changes; DC behavior is unchanged.

Affected Areas
- Enrollment Checker — Review page submit
- Self-Service Portal — sign-in callback page
- Self-Service Portal — initial dashboard load (and any authenticated page navigation while auth is being verified)

Test Cases

Happy Path
1. Enrollment Checker submit shows the interstitial. Open the CO Enrollment Checker, fill in at least one child, continue to the review page, and tap Submit. The form should disappear and the page should immediately show "Please wait..." with the info callout. When the response comes back, the page should automatically advance to the results screen.
2. Sign-in callback shows the interstitial. Sign in to the CO Self-Service Portal. The brief moment between completing sign-in at the identity provider and landing on the dashboard should display the "Please wait..." interstitial instead of any placeholder text or blank screen.
3. Dashboard initial load shows the interstitial. Once you land on the CO dashboard after sign-in, the "Please wait..." screen should remain visible until your household data is loaded; then the dashboard should render normally.
4. Refresh a protected page. Hard-refresh any signed-in CO portal page (dashboard, profile, cards, etc.). The "Please wait..." screen should appear briefly while the session is verified, then the page should render.

Edge Cases
1. Submission error. On the Enrollment Checker review page, force a submission failure (e.g., trigger the rate limit by submitting many times in a row). The interstitial should appear during the request and then disappear, returning to the review form with an error message at the top. The Submit button should be tappable again so the user can retry.
2. Sign-in failure. If the identity provider returns an error during sign-in callback, the interstitial should be replaced with a clear error alert (no infinite "Please wait..." loop).
3. Slow network. Throttle the network to "Slow 3G" in dev tools and submit on the review page or sign in to the portal. The interstitial should remain visible for the full duration of the request rather than flashing or disappearing prematurely.
4. Spanish language. Switch the language toggle to Español. The interstitial heading should read "Por favor, espera..." and the body should read "No salgas de la página. Estamos verificando si tenemos suficiente información."
5. Screen reader announcement. With VoiceOver or NVDA running, trigger any of the four entry points — the screen reader should announce the title and message politely (without interrupting the user) when the interstitial appears.

Regression Checks
1. DC portal users should NOT see the new interstitial. Sign in to the DC Self-Service Portal. The dashboard should load with its existing skeleton placeholder; the new "Please wait..." interstitial should not appear anywhere in the DC flow.
2. Enrollment Checker form back-button still works. Submit, let the interstitial appear, and let it finish. From the results page, use the browser back button — you should land back on the review form (not the interstitial).
3. Existing review page behavior. The review page should still let you edit each child, remove a child, add another child, and tap Back before you submit. None of those should be affected by the new loading state.
4. Auth redirect still works. Visit an authenticated portal URL while signed out (CO). You should still be redirected to the sign-in page — the interstitial should not block or replace that redirect.

Environment Notes
- Test on the CO deployment (STATE=co)
- The new interstitial is intentionally CO-only for this ticket. DC behavior is unchanged.
- Frontend-only change — no API, database, or config updates.

src/SEBT.Portal.Web/src/features/auth/components/AuthGuard/AuthGuard.tsx
```diff
@@ -31,8 +35,24 @@ export function AuthGuard({ children }: AuthGuardProps) {
     }
   }, [isHydrated, isAuthenticated, isLoading, router])
 
-  // Show nothing while hydrating, loading, or redirecting
-  if (!isHydrated || isLoading || !isAuthenticated) {
+  // While the auth-status fetch is in flight, CO shows an interstitial so the
+  // user knows the dashboard is still loading. Other states keep the existing
+  // blank-screen behavior until their content is wired up.
+  if (!isHydrated || isLoading) {
```
necampanini, line 40: liking the Interstitial component, but if the underlying requests stalls out, the user's never notified and is stuck on loading. maybe a timeout here that could flip to an error state would make it more obvious to the user

src/SEBT.Portal.Web/src/app/(public)/callback/page.tsx
```diff
@@ -110,23 +112,47 @@ export default function CallbackPage() {
...
+            heading={t('callbackSignInIssue')}
+          >
+            {errorDetail}
+          </Alert>
+        </div>
+      </div>
+    )
+  }
+
+  if (isCO) {
```
necampanini, line 135: this check and the wrapping `div ... grid-container maxw-tablet` appear a few times. may be worth pulling up into its own component, or extend the props on `<LoadingInterstitial/>` to take in classes?

src/SEBT.Portal.Web/src/features/household/components/DashboardContent/DashboardContent.tsx
```diff
@@ -51,7 +53,17 @@ export function DashboardContent() {
     return (
       <>
         {pageHeading}
```
necampanini, line 69: probably a nit but this could lead to having two `<h1>`s on the page at the same time. think it's allowed in WCAG standards but discouraged

packages/design-system/src/index.ts
```diff
@@ -23,6 +23,7 @@ export { TextLink } from './components/ui/TextLink'
 export type { TextLinkProps } from './components/ui/TextLink'
 export { SummaryBox } from './components/ui/SummaryBox'
 export type { SummaryBoxProps } from './components/ui/SummaryBox'
+export { LoadingInterstitial } from './components/ui/LoadingInterstitial'
```
necampanini, line 26: should we also export the `LoadingInterstitialProps` here too? following the pattern of other components here

src/SEBT.EnrollmentChecker.Web/src/app/review/page.tsx
```diff
@@ -39,6 +40,20 @@ export default function Page() {
     }
   }
 
+  if (isSubmitting) {
```
necampanini, line 43: maybe a larger topic to bring up since enrollment checker pages don't seem to have co-located test files for components

necampanini (changes requested):
solid PR, just requesting a few small changes

src/SEBT.Portal.Web/src/features/auth/components/AuthGuard/AuthGuard.tsx
adbergen: Good catch. I think the proper fix is to add a fetch timeout to fetchSession in AuthContext (AbortController + setTimeout) so we surface a real error signal, then AuthGuard reacts to that. Thinking this hsould be a follow up ticket.

packages/design-system/src/index.ts
adbergen: Addressed in [05059a3](https://github.com/codeforamerica/sebt-self-service-portal/pull/240/commits/05059a336e3c49e2464f695244616533bc8bb9ab)

src/SEBT.Portal.Web/src/features/household/components/DashboardContent/DashboardContent.tsx
adbergen: Addressed in [05059a3](https://github.com/codeforamerica/sebt-self-service-portal/pull/240/commits/05059a336e3c49e2464f695244616533bc8bb9ab)

src/SEBT.Portal.Web/src/app/(public)/callback/page.tsx
adbergen: Addressed in [05059a3](https://github.com/codeforamerica/sebt-self-service-portal/pull/240/commits/05059a336e3c49e2464f695244616533bc8bb9ab)

necampanini (approved):

PR #241, opened 2026-04-30
DC-348: Allow Amplitude SDK remote-config host in CSP

adbergen:
#### 🔗 Jira ticket
https://codeforamerica.atlassian.net/browse/DC-348

#### ✍️ Description
Amplitude analytics events were not flowing through in the CO dev environment because the browser SDK's initialization request to its remote-configuration endpoint (`https://sr-client-cfg.amplitude.com`) was blocked by our Content Security Policy. The Browser SDK v2 fetches this remote config on every init regardless of whether session replay is enabled, and a failed config fetch puts the SDK into a degraded state where event delivery is unreliable.

This change adds `https://sr-client-cfg.amplitude.com` to the `connect-src` directive alongside the existing `https://api2.amplitude.com` (events endpoint), gated by the same `hasAmplitude` check so the host is only injected when `NEXT_PUBLIC_AMPLITUDE_API_KEY` is set.

CO production was unaffected because `NEXT_PUBLIC_AMPLITUDE_API_KEY` isn't set there yet — `hasAmplitude` evaluates false, so the SDK never initializes and no blocked-fetch errors appear.

#### 🔗 Links to related PRs
N/A

#### ✅ Completion tasks

- [x] Added relevant tests
- [ ] Configuration changes:


adbergen:
# QA Summary — DC-348 Amplitude analytics not flowing in CO

## What Changed

The portal's Content Security Policy was blocking the Amplitude browser SDK's initialization request to its remote-configuration endpoint (`sr-client-cfg.amplitude.com`), which silently put the SDK into a degraded state and prevented events from being delivered. This change adds that host to the policy alongside the existing events host so the SDK can initialize cleanly. No user-visible UI change; verification is about analytics events landing in Amplitude.

## Affected Areas

- All pages — analytics events fired from any signed-in or signed-out page in the CO portal
- Browser console (no CSP error rows for `amplitude.com`)

## Test Cases

### Happy Path

1. **No CSP errors for Amplitude.** Open the CO portal (any page). Open browser DevTools → Console. There should be **no rows** that read "Refused to connect because it violates the document's Content Security Policy" referencing `sr-client-cfg.amplitude.com` or `api2.amplitude.com`. (Before this fix, those errors appeared on every page load in CO dev with the staging Amplitude key configured.)
2. **Page-view event reaches Amplitude.** Sign in to the CO portal and navigate through a few pages (dashboard → profile → cards). Open the Amplitude staging project's live events / user lookup view. Within ~30 seconds of each navigation, you should see a corresponding analytics event for your session.
3. **Sign-in event reaches Amplitude.** Sign out, then sign back in. Confirm the sign-in event appears in Amplitude's live event feed shortly afterward.
4. **CTA-click event reaches Amplitude.** Click a tracked dashboard CTA (e.g. "Update mailing address" or "Request a replacement card"). Confirm the corresponding click event lands in Amplitude.

### Edge Cases

1. **Amplitude key not configured.** On environments where `NEXT_PUBLIC_AMPLITUDE_API_KEY` is not set (e.g. CO production today), the SDK never initializes. The browser console should still be free of any Amplitude-related CSP errors — none should appear because the policy doesn't add the host when the key is absent.
2. **DC environment.** Sign in to the DC portal. Confirm DC behavior is unchanged (Amplitude is enabled or disabled per DC's own configuration; no new errors and existing analytics behavior is preserved).
3. **Slow / intermittent network.** Throttle to "Slow 3G" and reload a page. Amplitude's config request may take longer to complete, but it should not be blocked by the policy and events should eventually land.

### Regression Checks

1. **Other analytics still work.** Confirm Google Analytics 4, Mixpanel, and Google Tag Manager events continue to load and fire as before — these were already in the policy and should be unaffected.
2. **Socure ID-proofing still works.** Trigger the document-verification flow (CO step-up). Socure's embedded SDK and its Datadog telemetry should still load without CSP errors.
3. **PingOne sign-in still works.** Sign-in via the identity provider should still complete normally.
4. **No unrelated origins newly allowed.** Inspect the page response's `Content-Security-Policy` header (DevTools → Network → any HTML response → Headers). It should include `https://api2.amplitude.com https://sr-client-cfg.amplitude.com` together when Amplitude is enabled, and **no other new origins**.

## Environment Notes

- Test on the **CO dev** environment with `NEXT_PUBLIC_AMPLITUDE_API_KEY` set (Amplitude staging key). This is the environment where the bug was reproduced.
- Verifying the happy path requires access to the Amplitude project's live events / user lookup view.
- Frontend-only change — no API, database, or backend updates.

mtodd-cfa (approved):

PR #245, opened 2026-04-30
SECURITY: dependency cooldown config

robdiciuccio:
#### ✍️ Description
Supply chain security measures:
- Adds `pnpm` config for restricting installation of packages newer than 7 days
- Adds Dependabot configuration with 7d cooldown period
- ~Adds and enforces NuGet lockfiles and package source mapping~ (moved to follow-up task)

#### 🔗 Relevant docs
https://www.notion.so/cfa/Dependency-Management-Best-Practices-WIP-341373fd79b28043862ef8995b817f1d

#### ✅ Completion tasks

- [ ] Added relevant tests
- [ ] Configuration changes:


src/SEBT.Portal.Api/SEBT.Portal.Api.csproj
```diff
@@ -50,11 +50,7 @@
     <RemoveDir Directories="$(FrontendDist)" Condition="Exists('$(FrontendDist)')" />
   </Target>
 
-  <PropertyGroup>
-    <HomeDir Condition="'$(OS)' == 'Windows_NT'">$(USERPROFILE)</HomeDir>
```
robdiciuccio, line 54: If Windows support is important, we can address this separately.

robdiciuccio:
_EDIT: removing NuGet lockfiles for this PR. Will revisit later._

CI builds are failing for the following reasons (according to Claude). Let's discuss how to proceed here.

> **Root cause**
> 
> `Api.csproj:74-93` has a `<Choose>` block that picks `ProjectReference` if the sibling state-connector repo exists at the expected path, falling back to `PackageReference` otherwise. This is a local-dev convenience.
> 
> **The problem**: the lockfile's package set depends on which branch fires.
> 
> Locally (sibling exists): `ProjectReference` → `SEBT.Portal.StatesPlugins.Interfaces` does not appear as a package dependency in the lockfile.
> 
> In CI (sibling checked out to `<workspace>/state-connector/`, but the csproj checks `<parent-of-workspace>/state-connector/`, which doesn't exist): `PackageReference` → the package must appear in the lockfile.
> 
> **`--locked-mode` is fundamentally incompatible with the `Choose` block**: any committed lockfile is correct for one path and wrong for the other. There is no single lockfile that works for both.

jamesmblair (approved):

PR #273, opened 2026-05-04
DC-324(Feat): AES-256-GCM at-rest encryption for PII fields

noxmwalsh:
#### 🔗 Jira ticket

[DC-324](https://codeforamerica.atlassian.net/browse/DC-324)

#### ✍️ Description

Adds AES-256-GCM encryption at rest for reversible PII in the portal database, deterministic email lookup via `EmailHash`, and per-state operational toggles so Colorado can stay off while DC opts in when ready with a key.

This doesn't include the key rotations; this is designed to expect to a single key for now (which both states should be able to provide easily; they just need to update their ; we'll need to handle each state in a unique way, so they'll be tackled in a future.

Cryptography and configuration

- New `PiiEncryption` settings section with `EncryptAtRest` (default `false`), `RunStartupBackfill` (default `false`), `ActiveKeyId`, and a key ring (`KeyId` plus Base64 256-bit material).
- `IPiiSymmetricEncryption` with `PiiAesGcmSymmetricEncryption`: AES-GCM envelopes, per-write nonce, embedded key id, sentinel prefix (`sep-pii:v1:`), tamper detection via GCM tag.
- `PiiSymmetricEncryptionFactory` selects implementation at startup:
  - `PlaintextPiiSymmetricEncryption` when `EncryptAtRest` is false and no keys are configured (keyless mode for CO).
  - `ConditionalPiiSymmetricEncryption` wrapping AES-GCM when keys are present: encrypts writes only when `EncryptAtRest` is true; always decrypts existing envelopes on read.
- `IEmailLookupHasher` / `EmailLookupHasher`: HMAC fingerprint over normalized email for indexed equality lookups (domain-separated from `IdentifierHasher`).
- `PiiEncryptionSettingsValidator`: requires a valid key ring only when `EncryptAtRest` is true.
- `PiiEncryptionGuard`: in production, rejects placeholder dev keys only when `EncryptAtRest` is true (CO production with encryption off needs no PII encryption secrets).

Data model and migration

- EF migration `EncryptPiiAtRestColumnsAndEmailHash`: widens PII columns for ciphertext, adds `EmailHash`, drops unique index on plaintext `Email`, adds filtered unique index on `EmailHash`.
- Follow-up migration `AddUsersEmailLegacyLookupIndex`: non-unique filtered index on `Email` for transitional plaintext rows (`EmailHash` null).
- `Users.DateOfBirth` converted from SQL `date` to `nvarchar` via dynamic SQL in the migration batch.
- `UserEncryptedFieldMapper` centralizes encrypt/decrypt mapping for user entities.

Repositories and flows

- `DatabaseUserRepository`: reads/writes encrypted fields; lookup by `EmailHash` with transitional fallback for legacy plaintext rows; rejects duplicate emails when another user still stores plaintext for the same address.
- `DatabaseDocVerificationChallengeRepository`: encrypts ID-proofing challenge PII at rest.

Startup

- `PiiPlaintextEncryptionBackfill`: batched post-migration backfill of legacy plaintext to ciphertext and missing `EmailHash` values (idempotent; `ChangeTracker.Clear` after each batch).
- `PiiEncryptionStartupBackfill`: gates backfill on `RunStartupBackfill`; logs and swallows failures so startup is not blocked on transient DB issues.
- Wired from `Program.cs` after EF migrations, before seeding.

Seeding

- `DataSeeder` updated for hash-aware duplicate detection when bulk-adding users, including envelope rows missing `EmailHash`.

Configuration examples

- `appsettings.json`: both toggles `false` (safe default for all states).
- `appsettings.co.example.json`: explicit `false` / `false`.
- `appsettings.dc.example.json`: `EncryptAtRest` and `RunStartupBackfill` `true` with local dev placeholder keys.
- `appsettings.Development.example.json`: both toggles `false`.

Documentation

- `docs/adr/0015-pii-encryption-at-rest.md`: algorithm, key ring, toggles, rotation vs backfill, email lookup, logging constraints, backfill failure semantics.

Tests

- Crypto: `PiiAesGcmSymmetricEncryptionTests`, `ConditionalPiiSymmetricEncryptionTests`, `PlaintextPiiSymmetricEncryptionTests`, `PiiSymmetricEncryptionFactoryTests`, `PiiPlaintextEncryptionBackfillTests` (SQL Server)
- Config/guards/startup: `PiiEncryptionSettingsValidatorTests`, `PiiEncryptionGuardTests`, `PiiEncryptionStartupBackfillTests`
- Repos/seeding: `DatabaseUserRepositoryTests`, doc challenge repo tests, `DataSeederTests`, `DatabaseSeederTests`, `PortalDbContextTests`
- Shared helpers: `TestPortalCryptography`

All portal backend tests pass (1358 passed, 2 skipped).

#### 🔗 Links to related PRs

None yet. DC production enablement will need a follow-up for Tofu / AppConfig / Secrets Manager wiring when ops is ready to turn encryption on.

#### ✅ Completion tasks

- [x] Added relevant tests
- [ ] Configuration changes:
  - [x] If you're adding an appsetting, add it to the requisite example file, and update it in the AWS AppConfig

Deploy and ops notes

- CO: leave `EncryptAtRest` and `RunStartupBackfill` false. No PII encryption keys required.
- DC (when enabling): set `EncryptAtRest` true, configure real key id and key material in secrets, set `RunStartupBackfill` true for the initial rollout (can disable after backfill completes). Do not use `local-dev-primary` or the repo placeholder Base64 in production (`PiiEncryptionGuard`).
- On first DC enablement, run backfill on a single task or single-instance deploy if possible to avoid concurrent backfill races across multiple ECS tasks.
- After deploy: confirm migration and backfill logs; alert on backfill error logs per ADR. Spot-check login, profile read/write, and doc-verification flows for email, phone, DOB, and challenge data.
- Env var examples: `PiiEncryption__EncryptAtRest`, `PiiEncryption__RunStartupBackfill`, `PiiEncryption__ActiveKeyId`, `PiiEncryption__Keys__0__KeyId`, `PiiEncryption__Keys__0__KeyMaterialBase64`.

[DC-324]: https://codeforamerica.atlassian.net/browse/DC-324?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


src/SEBT.Portal.Core/AppSettings/PiiEncryptionSettings.cs
```diff
@@ -0,0 +1,70 @@
...
+            {
+                throw new InvalidOperationException($"Duplicate PII encryption KeyId '{id}'.");
+            }
+
+            dictionary[id] = Convert.FromBase64String(k.KeyMaterialBase64.Trim());
+        }
+
+        foreach (var (id, raw) in dictionary)
+        {
+            if (raw.Length is not (16 or 24 or 32))
```
necampanini: The new ADR in the PR says to use AES-256, should `16 or 24` be allowed in?

pkarman:
the ADR and algorithm pattern looks right to me.

The only suggestion I have is for planned expiration of PII. I.e. does it need to hang around forever in the db, or can it be deleted after some known period of time? While that may be out of scope for this particular PR, I can say that the only thing better than encrypting PII at rest is not storing PII at all.

src/SEBT.Portal.Infrastructure/Services/PiiPlaintextEncryptionBackfill.cs
```diff
@@ -0,0 +1,170 @@
...
+        await MigrateUsersLoopAsync(cancellationToken);
+        await MigrateDocVerificationLoopAsync(cancellationToken);
+    }
+
+    private async Task MigrateUsersLoopAsync(CancellationToken cancellationToken)
+    {
+        int migrated;
+        do
+        {
+            migrated = await MigrateOneUsersBatchAsync(cancellationToken);
```
necampanini, line 44: if this throws, would it rollback the whole batch? the catch in `Program.cs` doesn't discern between `PiiDecryptException` and a transient db error either

noxmwalsh: Addressed in a89b3897e882fd09872f2798278bd094e6da39a3

test/SEBT.Portal.Tests/Unit/UseCases/Auth/ValidateOtpCommandHandlerTests.cs
```diff
@@ -453,8 +453,7 @@ public async Task Handle_ShouldPassUserWithIalLevel_ToJwtTokenService()
             Email = command.Email,
             IalLevel = UserIalLevel.IAL1,
             IdProofingSessionId = "session-123",
-            IdProofingCompletedAt = null,
-            IdProofingExpiresAt = DateTime.UtcNow.AddYears(1)
```
noxmwalsh, line 457: IdProofingExpiresAt is not used anymore

necampanini (approved):

src/SEBT.Portal.Core/Services/IEmailLookupHasher.cs
```diff
@@ -0,0 +1,14 @@
+namespace SEBT.Portal.Core.Services;
+
+/// <summary>
+/// Computes a deterministic 64-character hex MAC for normalized email lookups (equality in SQL via persisted email-hash column).
+/// Separate from identifier hashing normalization rules used for SNAP/TANF/SSN.
+/// </summary>
+public interface IEmailLookupHasher
+{
+    /// <inheritdoc cref="NormalizeAndHash"/>
```
jamesmblair: This interface has no base interface, so no docs to inherit here.

src/SEBT.Portal.Infrastructure/Repositories/DatabaseUserRepository.cs
```diff
@@ -72,29 +85,43 @@ public async Task UpdateUserAsync(User user, CancellationToken cancellationToken
...
-        // OIDC users have null email — leave the DB value unchanged in that case.
-        if (user.Email != null)
+        if (NormalizeEmail(user.Email) != null)
         {
-            var normalizedEmail = NormalizeEmail(user.Email);
-            if (entity.Email != normalizedEmail)
+            var normalizedIncoming = NormalizeEmail(user.Email)!;
+            var envelopePrefix = PiiAesGcmSymmetricEncryption.EnvelopePrefix;
+            var plaintextDupe = await dbContext.Users.AnyAsync(
+                u =>
```
jamesmblair: In the migration, you removed the `IX_Users_Email` index, which could make this three clauses that are filtering on `Email` slow. We should probably keep the index?

src/SEBT.Portal.Infrastructure/Services/PiiPlaintextEncryptionBackfill.cs
```diff
@@ -0,0 +1,170 @@
...
+            challenge.UpdatedAt = DateTime.UtcNow;
+        }
+
+        if (batch.Count == 0)
+        {
+            return 0;
+        }
+
+        await _dbContext.SaveChangesAsync(cancellationToken);
+        _dbContext.ChangeTracker.Clear();
```
jamesmblair: Doesn't `SaveChangesAsync` automatically clear the change tracker?

src/SEBT.Portal.Core/AppSettings/PiiEncryptionSettings.cs
jamesmblair: Agreed on this

jamesmblair (changes requested):
Overall looks really good. Add a few minor comments.

src/SEBT.Portal.Infrastructure/Services/PiiPlaintextEncryptionBackfill.cs
noxmwalsh: Addressed in 96c742f2c9c55c6c446934e6a370865ca285e7ac

src/SEBT.Portal.Infrastructure/Repositories/DatabaseUserRepository.cs
noxmwalsh: Addressed in c5775742d12b6e2fe5b67e925a59602b71940b86

src/SEBT.Portal.Core/AppSettings/PiiEncryptionSettings.cs
noxmwalsh: Addressed in 1b3b0bdd414118b90c917a0e97c9caca5b76d009

src/SEBT.Portal.Core/Services/IEmailLookupHasher.cs
noxmwalsh: Addressed in 92f9c0d5a637f46738488af8ae78215235c9e3e9

jhernandez-cfa (approved):

src/SEBT.Portal.Api/Startup/PiiEncryptionStartupBackfill.cs
```diff
@@ -0,0 +1,66 @@
...
+/// </summary>
+public static class PiiEncryptionStartupBackfill
+{
+    internal const string SkippedBecauseDisabledMessage =
+        "PII ciphertext backfill skipped (PiiEncryption:RunStartupBackfill is false).";
+
+    /// <summary>Whether startup should invoke plaintext-to-ciphertext backfill after EF migrations.</summary>
+    public static bool ShouldRunStartupBackfill(PiiEncryptionSettings settings)
+    {
+        ArgumentNullException.ThrowIfNull(settings);
```
necampanini, line 20: if  `RunStartupBackfill` is set to true while `EncryptAtRest` is false, would the backfill write plaintext? if so the batch query keeps querying the same rows and the `do/while (migrated > 0)` loop would just spin

src/SEBT.Portal.Infrastructure/Repositories/DatabaseUserRepository.cs
```diff
@@ -2,29 +2,45 @@
...
             return null;
         }
 
-        var normalizedEmail = NormalizeEmail(email);
         var entity = await dbContext.Users
             .AsNoTracking()
-            .FirstOrDefaultAsync(u => u.Email == normalizedEmail, cancellationToken);
+            .FirstOrDefaultAsync(
+                u =>
+                    u.EmailHash == lookupHash ||
```
necampanini: there may be an edge case for partially encrypted rows where email is encrypted but the email hash is null so it falls through this predicate.  we'd get a duplicate row from that, right?

necampanini (changes requested):
looking great! just the 2 callouts

src/SEBT.Portal.Infrastructure/Repositories/DatabaseUserRepository.cs
noxmwalsh: Addressed in 750b4931c2038c5c0a3fbec9fd25b578760ba069

src/SEBT.Portal.Api/Startup/PiiEncryptionStartupBackfill.cs
noxmwalsh: Addressed in 750b4931c2038c5c0a3fbec9fd25b578760ba069

necampanini (approved):

PR #277, opened 2026-05-05
DC-148: load Mixpanel via the official boot snippet

adbergen:
#### 🔗 Jira ticket
https://codeforamerica.atlassian.net/browse/DC-148

#### ✍️ Description
Fixes the runtime error \`Mixpanel error: \"mixpanel\" object not initialized\` reported in dev across CO and DC.

**Root cause**: \`MixpanelAnalytics\` was loading the Mixpanel CDN library directly with \`<Script src=\"…mixpanel-2-latest.min.js\">\`. The library by itself never assigns \`window.mixpanel\` — that's the job of Mixpanel's official boot snippet, which pre-defines a queue stub on \`window.mixpanel\` and then asynchronously loads the lib (which upgrades the stub on execution). Skipping the snippet meant \`window.mixpanel\` stayed undefined, the bridge's gate \`if (!mp?.init || !mp?.track)\` silently no-op'd, and any later \`mixpanel.*\` call fell into the lib's \"not initialized\" complaint.

**Fix**:
- \`MixpanelAnalytics.tsx\` now inlines Mixpanel's official boot snippet via \`<Script dangerouslySetInnerHTML>\` (CSP-friendly with the existing \`nonce\` prop). The snippet itself loads the CDN script asynchronously, so the explicit \`src=\"…\"\` is gone.
- \`mixpanel-bridge.ts\` gate relaxes from \`!mp?.init || !mp?.track\` to \`!mp?.init\`. The snippet defines \`init\` immediately on the queue stub but only attaches \`track\` (and the rest of the API surface) the first time \`init()\` runs — checking for \`track\` up-front would always fail and never bootstrap the SDK.
- New \`MixpanelAnalytics.test.tsx\` locks the wiring against regression: asserts the inline snippet contains the queue-stub assignment and the CDN URL, asserts \`initMixpanelBridge\` is called only after \`onReady\`, and asserts the token is forwarded.

#### 🔗 Links to related PRs
None.

#### ✅ Completion tasks
- [x] Added relevant tests (3 new + full portal suite 748/748, analytics package 92/92)
- [ ] Configuration changes:

#### Validation
Local repro: dev server with \`NEXT_PUBLIC_MIXPANEL_TOKEN\` set → before this PR, the console emits \"mixpanel object not initialized\" on every page load. After this PR, the console is clean and Mixpanel's debug view shows incoming events.

### DIFF
--- a/packages/analytics/src/mixpanel-bridge.ts
+++ b/packages/analytics/src/mixpanel-bridge.ts
@@ -83,7 +83,11 @@ function attachBridge(dl: DataLayerRoot): () => void {
 
 export function initMixpanelBridge(token: string): () => void {
   const mp = window.mixpanel
-  if (!mp?.init || !mp?.track) {
+  // Only `init` is required up-front. The official Mixpanel boot snippet
+  // defines `init` immediately on the queue stub but only attaches `track`
+  // (and the rest of the API surface) the first time `init()` is called —
+  // checking for `track` here would always fail and never bootstrap the SDK.
+  if (!mp?.init) {
     return () => {}
   }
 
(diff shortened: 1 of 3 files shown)


necampanini (approved):
looks good!

PR #284, opened 2026-05-06
DC-359 Update: Show suggestion screen when Smarty has a suggestion

noxmwalsh:
https://github.com/user-attachments/assets/c45f9135-f0f5-4773-9de0-b5435e167508

#### 🔗 Jira ticket

[DC-359](https://codeforamerica.atlassian.net/browse/DC-359)

#### ✍️ Description

This work improves the mailing address update flow when Smarty returns a normalized address that meaningfully differs from what the user typed.

Backend (`UpdateAddressCommandHandler`)

- After Smarty/state validation, compare the requested address to the normalized address (trim + case-insensitive per field).
- If they are not equivalent, return `AddressValidationResult.Suggestion` with reason `suggested` so the client can route to the suggestion UI (not only the DC 30-character abbreviation path).
- Case-only differences do not trigger a suggestion.
- Unchanged behavior for paths that already returned suggestion/invalid from state-specific validation.

Frontend (`SuggestedAddress`)

- When the user chooses Suggested address, treat API responses with `status: "valid"` or `status: "suggestion"` as success for navigation (covers cases where a second normalization still returns suggestion).
- Prefer `normalizedAddress` from the response when present; otherwise fall back to `suggestedAddress` before updating flow context and continuing.

Tests

- `UpdateAddressCommandHandlerTests`: coverage for suggestion when normalization differs, no suggestion when only casing differs, and adjustment of expectations where normalization previously proceeded without suggestion.
- `SuggestedAddress.test.tsx`: new test for continuation when the API returns `suggestion` after accepting the suggested address.

#### 🔗 Links to related PRs

[DC-connector](https://github.com/codeforamerica/sebt-self-service-portal-dc-connector/pull/35)

#### ✅ Completion tasks

- [x] Added relevant tests
- [ ] Configuration changes:
  - [ ] If appsettings are changed, update in AWS AppConfig

Suggested QA

1. DC user with `canUpdateAddress`, enter an address Smarty will normalize (e.g. long MLK street name -> abbreviated). 
3. Confirm `422` with `status: "suggestion"` on first submit and suggestion screen appears.
4. Choose suggested vs entered and confirm downstream screens and `PUT /api/household/address` behavior match product expectations (including when backend returns `valid` vs another `suggestion`).

[DC-359]: https://codeforamerica.atlassian.net/browse/DC-359?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/SEBT.Portal.UseCases/Household/UpdateAddress/UpdateAddressCommandHandler.cs
+++ b/src/SEBT.Portal.UseCases/Household/UpdateAddress/UpdateAddressCommandHandler.cs
@@ -88,18 +88,28 @@ public class UpdateAddressCommandHandler(
 
         // Smarty's USPS-long canonical street can exceed DC's connector limit while the user's typed
         // street still fits. Abbreviated suggestions must not loop forever when the user explicitly
-        // chooses "address you entered" — validate and persist those lines instead.
-        var abbreviatedOnlyBlocksPersist =
-            command.AcceptEnteredAddress
-            && !stateValidationOnNormalized.IsValid
-            && stateValidationOnNormalized.Reason == "abbreviated";
-
-        if (!stateValidationOnNormalized.IsValid && !abbreviatedOnlyBlocksPersist)
+        // chooses "address you entered" - validate and persist those lines instead.
+        var abbreviatedOnlyBlocksPersist = false;
+        if (!stateValidationOnNormalized.IsValid)
         {
-            logger.LogInformation(
-                "Address failed state-specific validation: {Reason}",
-                stateValidationOnNormalized.Reason);
-            return Result<AddressValidationResult>.Success(stateValidationOnNormalized);
+            // Only abbreviated-normalization mismatch may proceed; entered lines are validated below before persist.
+            if (stateValidationOnNormalized.Reason != "abbreviated")
+            {
+                logger.LogInformation(
+                    "Address failed state-specific validation: {Reason}",
+                    stateValidationOnNormalized.Reason);
+                return Result<AddressValidationResult>.Success(stateValidationOnNormalized);
+            }
+
+            if (!command.AcceptEnteredAddress)
+            {
+                logger.LogInformation(
+                    "Address failed state-specific validation: {Reason}",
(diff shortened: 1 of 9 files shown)


src/SEBT.Portal.UseCases/Household/UpdateAddress/UpdateAddressCommandHandler.cs
```diff
@@ -215,4 +232,20 @@ public async Task<Result<AddressValidationResult>> Handle(UpdateAddressCommand c
                 "Address update service is temporarily unavailable.");
         }
     }
+
+    private static bool AddressesEquivalent(Address left, Address right)
```
spokenbird: I think this is essentially repeating what is already handled by the AddressNormalizationHelper.AddressesEqualLoose ? 

I think this logic could be replaced by capturing success.Value.WasCorrected in the switch above, and then using it here, so something like:

```
if (wasCorrected)
  {                
      logger.LogInformation("Address normalization produced a suggested alternative");                                                          
      return Result<AddressValidationResult>.Success(                                 
          AddressValidationResult.Suggestion(normalizedAddress!, "suggested"));                                                                 
  } 
```
Could replace the helpers and the logic on 96 to 109 since the success result above has already undergone that check and that way we keep `AddressNormalizationHelper.AddressesEqualLoose` as the ultimate decider of if smarty changed something.

spokenbird (changes requested):
Requesting changes to reuse the existing address comparison pattern

src/SEBT.Portal.UseCases/Household/UpdateAddress/UpdateAddressCommandHandler.cs
noxmwalsh: Addressed in 0fceaf6798f556452a7dc7cfefa55d95f57638b9

spokenbird (approved):

PR #288, opened 2026-05-07
DC-314(chore): i18n integrity gate (pre-commit + prebuild) with baseline ratchet and fix copy

adbergen:
#### 🔗 Jira ticket
N/A — tech-debt initiative surfaced from a discussion about copy-pipeline drift. Worth filing a follow-up to track the cleanup queue.

#### ✍️ Description
Adds a deterministic check that catches what the CSV → JSON pipeline can't:

- code calls \`t('foo')\` but the sheet has no \`foo\` row → **missing**
- code calls \`t('foo')\`, sheet has the row, cell is blank → **empty**
- both of the above with a string fallback hiding the gap → **masked** (warn)
- sheet ships a key nothing renders → **orphan** (warn)

Cross-checks every \`t()\` call site in \`src/SEBT.Portal.Web/src/\` against the generated locale JSON for every state × locale (co/dc × en/es). Implementation in [\`packages/design-system/content/scripts/audit-i18n.js\`](packages/design-system/content/scripts/audit-i18n.js) — TS-AST based so it correctly handles aliased bindings like \`const { t: tCommon } = useTranslation('common')\` and skips dynamic keys.

#### Where it gates

| Location | When it fires |
|---|---|
| Pre-commit hook | Any commit that touches \`src/SEBT.Portal.Web/\` or \`packages/design-system/content/\`. Fails fast before the slower frontend build/test steps. |
| \`prebuild\` (\`pnpm build\`) | CI safety net — production builds + deploys fail on regressions even if the hook was bypassed. |
| Manual: \`pnpm lint:i18n\` | Ad-hoc check. |

#### The baseline ratchet (the only reason this can ship today)

The audit found **97 unmasked errors**, **550 masked errors**, and **353 orphan keys** at HEAD of \`main\`. \`docs/i18n-audit-baseline.json\` snapshots the 97 unmasked entries so the gate enforces *no new regressions* without breaking the existing build. The baseline is one-way: shrink with deliberate cleanup work, never grow.

To clear an entry once the sheet is fixed:
\`\`\`sh
cd src/SEBT.Portal.Web
pnpm copy:generate                # pulls in updated CSV
pnpm lint:i18n:update-baseline    # writes a smaller baseline
git add ../../docs/i18n-audit-baseline.json
\`\`\`

The diff against the baseline reads like a release note — every removed entry is one missing/empty user-facing string fixed.

#### What it does NOT catch (follow-ups)

- Hardcoded JSX text (\`<p>Hello</p>\`) that bypasses i18next entirely. Tracked separately — would use \`eslint-plugin-i18next/no-literal-string\` with a curated allowlist.
- Backend C# strings that surface to the UI.
- Dynamic keys (\`\`t(\`prefix.\${name}\`)\`\`) — silently skipped.
- Pluralization / interpolation correctness.

#### Verification

- New canary file with a fake \`t('common:this_key_does_not_exist_in_csv')\` call → audit reports 4 errors (en×es × co×dc), exits 1. ✅
- Same canary with the call site removed → audit exits 0 with the 97 grandfathered baseline entries silenced. ✅
- Hook syntax-check (\`bash -n\`) passes. ✅
- Pre-commit hook ran on this PR's own commits and passed (758/758 frontend tests still green). ✅

#### 🔗 Links to related PRs
N/A

#### ✅ Completion tasks
- [x] Added relevant tests (smoke-tested via canary; hook ran on this PR's commits)
- [x] Meets acceptance criteria
- [ ] Configuration changes:


src/SEBT.Portal.Web/src/app/(public)/login/id-proofing/off-boarding/page.tsx
```diff
@@ -83,9 +83,9 @@ export default function OffBoardingPage() {
     body = t('body1')
     // TODO: Use t('action1') once key is available in dc.csv
     contactLabel = tCommon('linkContactUs')
-    applyBody = t('body2', '') || undefined
```
necampanini, line 86: at least my version of the copy, some of these would return key strings like `"action"` and `"body2"` instead of being empty via ` || undefined` 

a few new config props would be a quick fix in the `i18n.ts` file

```
i18n.use(initReactI18next).init({
...
// make t() falsy for misses so `t(key) || fallback` works.
returnEmptyString: false,
returnNull: false,
parseMissingKeyHandler: (_key, defaultValue) => defaultValue ?? ''
...
})
```

.husky/pre-commit
```diff
@@ -85,6 +103,36 @@ if [ "$FRONTEND_CHANGES" = "true" ]; then
...
+
+  # Regenerate locales first — a CSV change must produce fresh JSON before the
+  # audit cross-checks code against it. The generator is hash-cached so the
+  # no-op path is fast.
+  pnpm --silent copy:generate || {
+    log_error "Locale generation failed"
+    exit 1
+  }
+
+  pnpm --silent lint:i18n || {
```
necampanini, line 126: not blocking but possibly helpful in debugging non-local environments: could we add something to capture the output from a failed run to see where things break? or running without the `--silent` if its not too noisy

packages/design-system/content/scripts/audit-i18n.js
```diff
@@ -0,0 +1,539 @@
...
+  // — Key mismatches (code-side fix, ship in the next PR)
+  lines.push(`## 🔧 Code-side fixes — key mismatch`)
+  lines.push(``)
+  lines.push(`Engineer's fallback string is identical to a value already in the sheet at a different key. Most likely the call site is calling the wrong key (typo, copy-paste from another component, or a key that got renamed in the sheet but not in code). These are fixed in the **code**, not the sheet — change the \`t()\` call to use the matched key.`)
+  lines.push(``)
+  if (keyMismatches.length) {
+    lines.push(`| Current call | Likely correct key | Fallback string | Where |`)
+    lines.push(`|---|---|---|---|`)
+    for (const m of keyMismatches.slice(0, 200)) {
+      const safe = m.fallback.replace(/\|/g, '\\|').replace(/\n/g, ' ')
```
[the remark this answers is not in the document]
necampanini: suggested fix seems sound

necampanini (changes requested):
overall looking great, and the truthy fallback issue in the offboarding page may be out-of-date copy csv issue.

other than that, the github security alert is small but worth fixing as well

PR #291, opened 2026-05-08
DC-295: hashed SEBT App ID + portal_id in analytics data layer

adbergen:
#### 🔗 Jira ticket
https://codeforamerica.atlassian.net/browse/DC-295

#### ✍️ Description
Surfaces two new analytics fields so CO/CfA can join Mixpanel and Amplitude exports back to state program data without leaking raw identifiers to vendor tooling:

- **`digitalData.user.hashed_app_id`** — HMAC-SHA256 of the SEBT App ID, lowercase hex. CO-only (gated on `STATE=co` server-side); other states never see the field.
- **`digitalData.user.portal_id`** — stable, non-PII portal user UUID (`User.Id`, Guid v7). All states.

Both fields are emitted on the dashboard's `household_result` event with the existing `['default', 'analytics']` scope, so the existing Mixpanel + Amplitude bridges pick them up via the standard `digitalData.user.*` flow without bridge changes.

#### Backend

- New method `IIdentifierHasher.HashForAnalytics(string?)` ([IdentifierHasher.cs](src/SEBT.Portal.Infrastructure/Services/IdentifierHasher.cs)) that bypasses storage-side normalization (no trim/dash-strip) and returns lowercase hex. The storage `Hash()` method is unchanged so existing PII hashing keeps working.
- `AuthorizationStatusResponse` now carries `UserId` (Guid). `GET /api/auth/status` exposes the portal's own user UUID for analytics correlation.
- `HouseholdDataResponse` now carries `HashedAppId` (string?). `HouseholdController` computes it from the first `Application.ApplicationNumber`, gated on `STATE=co`. Mapper stays a pure projection.
- If `IdentifierHasher:SecretKey` is missing the app fails fast at startup (existing `IdentifierHasherGuard`), so `HashedAppId` is never silently dropped.

#### Frontend

- `SessionInfo.userId` flows from the auth-status response into `useUserDataSync`, which sets `setUserData('portal_id', userId, ['default', 'analytics'])` whenever the session has a UUID.
- `HouseholdDataSchema.hashedAppId` is the new (nullable) field. `DashboardContent` sets `setUserData('hashed_app_id', ...)` when the API returns one. When absent we don't touch the data layer at all so analytics never sees a stale digest.

#### Spec + reference script

[docs/analytics/hashed-sebt-app-id.md](docs/analytics/hashed-sebt-app-id.md) is the contract for downstream consumers. It documents the algorithm, encoding, normalization rules, configuration, and a concrete test vector. [docs/analytics/scripts/hash_sebt_app_id.py](docs/analytics/scripts/hash_sebt_app_id.py) is a stdlib-only Python reference that produces the same digest. The backend test `IdentifierHasherTests.HashForAnalytics_TestVector_MatchesPublishedReference` asserts the documented vector, so any algorithm change fails CI rather than silently desynchronizing external pipelines.

| Field | Value |
|---|---|
| Secret | `TestVectorSecret_AtLeast32Bytes_!!!!` |
| Input | `APP-2024-0001` |
| Digest | `ca383d90647e371547d6e66297cda8089b81fc1c5cb30da6cfcbdf744d9e2861` |

Verified: `python3 docs/analytics/scripts/hash_sebt_app_id.py 'APP-2024-0001' 'TestVectorSecret_AtLeast32Bytes_!!!!'` → matches.

#### Tests

Backend (1110/1110): 6 new hasher tests (lowercase hex, idempotence, no-normalization, null/whitespace, secret-mismatch, test-vector parity); 2 mapper tests (pass-through, default-null); HouseholdControllerTests updated to thread the new injected hasher.

Frontend (771/771): new `useUserDataSync.test.tsx` covers `portal_id` emission + suppression; new `hashed_app_id` test bucket on `DashboardContent.test.tsx` covers presence + non-CO absence; existing session-shape tests updated for the new `userId` field.

#### 🔗 Links to related PRs
None — single-repo feature.

#### ✅ Completion tasks
- [x] Added relevant tests (8 new backend, 4 new frontend; suites green)
- [ ] Configuration changes:
  - [ ] If new environment variables are added, update in Tofu — `IdentifierHasher__SecretKey` already exists for storage hashing; reused here. Confirm CO env has it set.
  - [ ] If new environment secrets are added, update in Tofu and set in AWS Secret Manager — same secret as above.
  - [ ] If you're adding an appsetting, add it to the requisite example file, and update it in the AWS AppConfig — n/a (reuses existing key)


src/SEBT.Portal.Infrastructure/Services/IdentifierHasher.cs
```diff
@@ -59,6 +59,18 @@ public bool Matches(string? plaintext, string? storedHash)
...
 
+    /// <inheritdoc />
+    public string? HashForAnalytics(string? plaintext)
+    {
+        if (string.IsNullOrWhiteSpace(plaintext))
+        {
+            return null;
+        }
+
+        var hash = HMACSHA256.HashData(_keyBytes, Encoding.UTF8.GetBytes(plaintext));
```
necampanini: `_keyBytes` looks like its the same key used in cooldown lookup hashes for card replacements. if we rotate this for analytics, wouldn't invalidate any existing card replacement cooldown row?

may be worth a separate `IdentifierHasher:AnalyticsSecretKey`

src/SEBT.Portal.Api/Controllers/Household/HouseholdController.cs
```diff
@@ -57,6 +59,29 @@ public async Task<IActionResult> GetHouseholdData(
...
+    /// Resolves the analytics-side hashed SEBT App ID for the household. CO-only
+    /// today (gated on the active state so DC payloads stay unchanged). Returns
+    /// null when no application carries an ApplicationNumber (e.g. auto-issued
+    /// SummerEbt cases). The frontend treats null as "do not emit".
+    /// </summary>
+    private static string? ResolveHashedAppId(
+        Core.Models.Household.HouseholdData data,
+        IIdentifierHasher identifierHasher)
+    {
+        var state = Environment.GetEnvironmentVariable("STATE");
```
necampanini: I don't think we have some injected settings that our backend could use to determine what state currently is, but it would be nice for things like this.

as is, we can't write a test against this asserting it's Colorado without manipulating process level variables.

not blocking but thought I'd bring it up

src/SEBT.Portal.Web/src/features/household/api/schema.ts
```diff
@@ -278,7 +278,11 @@ export const HouseholdDataSchema = z.object({
   benefitIssuanceType: IssuanceTypeSchema.nullable().optional(),
   allowedActions: AllowedActionsSchema.nullable().optional(),
   // Missing/null preprocess to Unknown so analytics never collapse broken payloads into NonCoLoaded (PR #208).
-  coLoadedCohort: CoLoadedCohortSchema
+  coLoadedCohort: CoLoadedCohortSchema,
+  // HMAC-SHA256 digest of the SEBT App ID (lowercase hex). Backend emits this
+  // only for states configured to surface it (CO today). Null otherwise.
+  // See docs/analytics/hashed-sebt-app-id.md.
+  hashedAppId: z.string().nullable().optional()
```
necampanini: while the C# backend rejects whitespace, I think this here would allow a `" "` whitespace to get through, so this would be a defensive change in case the backend changes in the future

src/SEBT.Portal.Api/Controllers/Household/HouseholdController.cs
```diff
@@ -57,6 +59,29 @@ public async Task<IActionResult> GetHouseholdData(
...
+        if (!string.Equals(state, "co", StringComparison.OrdinalIgnoreCase))
+        {
+            return null;
+        }
+
+        var applicationNumber = data.Applications
+            .Select(a => a.ApplicationNumber)
+            .FirstOrDefault(n => !string.IsNullOrWhiteSpace(n));
```
necampanini: are Colorado households guaranteed to have only one application? if not, the `hashed_app_id` could possibly change across page loads.

src/SEBT.Portal.Api/Controllers/Auth/AuthController.cs
```diff
@@ -52,6 +52,7 @@ public IActionResult GetAuthorizationStatus()
 
         return Ok(new AuthorizationStatusResponse(
             IsAuthorized: true,
+            UserId: userId,
```
necampanini, line 55: Might be worth adding a test or two around this new prop to defensively lock in the shape that the JWT claim expects.

necampanini (changes requested):
lot of good work here, just a few small changes I'd address

necampanini (approved):

PR #293, opened 2026-05-08
Fix images in static build enrollment app with BASE_PATH

### LINKED ISSUE #292: CVE-2026-44503
https://advisories.gitlab.com/nuget/microsoft.kiota.abstractions/CVE-2026-44503/

pkarman:
#### ✍️ Description

- fixes #292 
- adds `basePath` support to the static enrollment build, and uses dynamic `src` attributes for all images in the enrollment app
- 
#### ✅ Completion tasks

- [ ] Added relevant tests
- [ ] Configuration changes:


src/SEBT.EnrollmentChecker.Web/next.config.ts
```diff
@@ -13,12 +14,14 @@ if (process.env.BUILD_STANDALONE === 'true' && process.env.BUILD_STATIC === 'tru
...
   reactCompiler: true,
+  basePath: basePath,
   transpilePackages: ['@sebt/design-system'],
   // Note: react-i18next is NOT in serverExternalPackages here (unlike the portal).
   // Instead, layout.tsx uses direct imports from @sebt/design-system subpaths to
   // avoid pulling react-i18next into the RSC server bundle via the barrel export.
   env: {
-    NEXT_PUBLIC_STATE: state
+    NEXT_PUBLIC_STATE: state,
+    NEXT_BASE_PATH: basePath,
```
necampanini: `NEXT_BASE_PATH` is defined here but I'm not seeing it in the `Dockerfile.ssg`, would that be necessary?

necampanini (commented):
looks good, I just have that one open question about `NEXT_BASE_PATH` being defined in the `next.config.ts` but not referred to in the docker file

src/SEBT.EnrollmentChecker.Web/next.config.ts
pkarman: it isn't necessary for CO because we don't use the Dockerfile - we build the app statically and deploy to an S3 bucket.

I imagine for dynamic enrollment app it would be useful, so I can add it there for other users for completeness.

necampanini (approved):

src/SEBT.EnrollmentChecker.Web/next.config.ts
pkarman: added in 05acd20 - I am using `BASE_PATH` as the actual build-time name, and then setting it in the app env as `NEXT_BASE_PATH` which seems to be the convention. Unless maybe it should be `NEXT_PUBLIC_BASE_PATH` to indicate that the value is visible to users?

pkarman: 0037241 renames env var for consistency internally.

spokenbird:
@pkarman confirming the additional updates look good, if that's everything I can get this merged?

PR #298, opened 2026-05-11
DC-335: Clear EC cache on landing and after 15min idle

adbergen:
#### 🔗 Jira ticket
https://codeforamerica.atlassian.net/browse/DC-335

#### ✍️ Description

Two changes to the Enrollment Checker to limit child-info exposure on shared devices:

1. **Landing page mounts → cache cleared.** `LandingPage` now calls `clearState()` on mount. Clicking the logo (hardcoded `<Link href="/">` in the shared `Header`) routes here, so any deep-page → logo navigation now wipes children and `sessionStorage['enrollmentState']`. Direct URL navigation to `/` from a deep page is also cleared — the home screen is treated as a fresh-start screen.

2. **15-minute idle timeout.** New `useIdleTimeout(onTimeout, timeoutMs)` hook listens to `mousedown` / `keydown` / `touchstart` / `scroll` / `visibilitychange`. The timer resets on any activity (and when the tab returns to the foreground; hiding a tab does **not** reset, so a backgrounded tab still times out). Wired inside `EnrollmentProvider` so 15 minutes of inactivity calls `clearState()` and removes child info from memory + sessionStorage. Threshold lives on `IDLE_TIMEOUT_MS` next to `MAX_CHILDREN` in `EnrollmentContext.tsx`.

The shared design-system `Header` was intentionally **not** modified — keeping the "fresh-start on landing" semantic in the consuming app avoids a cross-app prop plumbing change.

**Tests**
- `EnrollmentContext.test.tsx`: two new fake-timer tests — clears after `IDLE_TIMEOUT_MS`; resets on `keydown` then clears on the next full window.
- `LandingPage.test.tsx`: existing tests now wrap in `EnrollmentProvider` (component depends on the context); new test seeds `sessionStorage` with a child and asserts it's wiped on mount.

EC suite: 100 → 103 passing.

#### 🔗 Links to related PRs
N/A (single-repo change)

#### ✅ Completion tasks

- [x] Added relevant tests
- [ ] Configuration changes:


adbergen:
# QA Summary — DC-335

## What Changed

The Enrollment Checker now clears any previously entered child information in two situations:

1. **When the user returns to the home screen** (by clicking the state logo in the top-left of the header, or by navigating directly to the home URL).
2. **After 15 minutes of inactivity** anywhere in the Enrollment Checker (no clicks, taps, key presses, or scrolling).

This reduces exposure of children's personal info on shared/public devices.

## Affected Areas

- Enrollment Checker — Home / Landing page
- Enrollment Checker — Disclaimer page
- Enrollment Checker — Child form / Add a child page
- Enrollment Checker — Review children page
- Enrollment Checker — Results page
- Header / logo (top-left of every Enrollment Checker page)

## Test Cases

### Happy Path

- [ ] **Logo click clears children.**
  1. Start on the Enrollment Checker home page.
  2. Click the primary action button (e.g. "Apply now") and continue through the disclaimer.
  3. Add at least one child (first name, last name, date of birth).
  4. Continue to the Review page and confirm the child appears in the list.
  5. Click the state logo in the top-left of the header.
  6. **Expected:** You land back on the home page. Now click through to the child-form / review pages again — the previously entered child should be gone, and you should see an empty state.

- [ ] **Direct navigation to home clears children.**
  1. Add a child and reach the Review page.
  2. Replace the URL in the address bar with the bare Enrollment Checker home URL and press Enter.
  3. **Expected:** Home page loads, and the previously entered child is no longer in the cache (verify by walking back through the flow).

- [ ] **15-minute inactivity timeout clears children.**
  1. Add at least one child and stop on the Review page.
  2. Leave the page completely idle for 15 minutes — do not click, scroll, type, or move the mouse over the window. Leaving the tab focused but untouched is fine.
  3. After 15 minutes, navigate back through the flow.
  4. **Expected:** The previously entered child is gone; the children list is empty.

### Edge Cases

- [ ] **Activity resets the idle timer.**
  1. Add a child and reach the Review page.
  2. Wait ~10 minutes, then click anywhere on the page (or press a key).
  3. Wait another 10 minutes after that interaction.
  4. **Expected:** Child is still present (the second 10-minute window is well under the 15-minute timeout that restarted at the interaction).

- [ ] **Backgrounded tab still times out.**
  1. Add a child and reach the Review page.
  2. Switch to a different browser tab or application and leave the Enrollment Checker tab untouched for 15+ minutes.
  3. Return to the Enrollment Checker tab.
  4. **Expected:** Child has been cleared (a tab being in the background does not protect the cache from timing out).

- [ ] **Returning to the foreground resets the timer.**
  1. Add a child, then switch to another tab for ~5 minutes (well under 15).
  2. Return to the Enrollment Checker tab.
  3. Wait 14 more minutes from the moment you returned.
  4. **Expected:** Child is still present (returning to foreground resets the clock).

- [ ] **Refreshing the home page is a no-op.**
  1. Land on the home page with no children entered.
  2. Refresh the browser.
  3. **Expected:** Page loads normally — no errors, behaves identically to a first visit.

### Regression Checks

- [ ] Add child flow still works end-to-end: home → disclaimer → add child → review → submit → results.
- [ ] Editing a child on the review page still saves changes correctly.
- [ ] Removing a child from the review page still works.
- [ ] Adding multiple children (up to the maximum of 20) still works.
- [ ] Language toggle (English / Spanish) still works on the home page.
- [ ] FAQ accordion on the home page still expands and collapses.

## Test Credentials

Not applicable — the Enrollment Checker is a public, unauthenticated tool. No sign-in is required.

## Environment Notes

- Test on the Enrollment Checker app (separate from the SEBT Portal sign-in app). Both states (CO and DC) ship the same idle-timeout behavior.
- The 15-minute timeout uses real wall-clock time and cannot be shortened via a flag for testing — please allow the full 15 minutes for the inactivity case rather than trying to fast-forward.
- No environment variables, secrets, or app-settings changes were made for this work — nothing to update in Tofu, AWS Secret Manager, or AppConfig.

src/SEBT.EnrollmentChecker.Web/src/features/enrollment/context/useIdleTimeout.ts
```diff
@@ -0,0 +1,42 @@
+'use client'
+
+import { useEffect, useRef } from 'react'
+
+const ACTIVITY_EVENTS = ['mousedown', 'keydown', 'touchstart', 'scroll'] as const
+
+// Resets the timer when the tab returns to the foreground but not when it
+// hides — a backgrounded tab should still time out.
+export function useIdleTimeout(onTimeout: () => void, timeoutMs: number): void {
```
jamesmblair, line 9: Note that we have something similar [here](https://github.com/codeforamerica/sebt-self-service-portal/blob/main/src/SEBT.Portal.Web/src/features/auth/idle/useUserActivity.ts) in the Portal web app. 

Maybe this is fine for now? But we should consider aligning for consistency.

src/SEBT.EnrollmentChecker.Web/src/features/enrollment/context/EnrollmentContext.tsx
```diff
@@ -119,6 +123,13 @@ export function EnrollmentProvider({ children }: { children: ReactNode }) {
     })
   }
 
+  function clearState() {
+    if (typeof window !== 'undefined') sessionStorage.removeItem(STORAGE_KEY)
+    setState(initialState)
+  }
+
+  useIdleTimeout(clearState, IDLE_TIMEOUT_MS)
```
jamesmblair: When clearing state on idle timeout, should we route the user back to the landing page?

jamesmblair (changes requested):

src/SEBT.EnrollmentChecker.Web/src/features/enrollment/context/useIdleTimeout.ts
adbergen: The portal hook tracks the last activity timestamp and lets a separate scheduler poll it to decide when to refresh the OAuth token. This one just needs to fire a callback when the user goes idle, so the debounced timer shape ends up simpler.

jamesmblair: Yep, agreed on that part. My alignment concern is mainly around a unified definition of idle/active. Consistent event tracking. This is fine for now though.

jamesmblair (approved):

PR #299, opened 2026-05-11
DC-393: forward page_load events to Amplitude

adbergen:
#### 🔗 Jira ticket
https://codeforamerica.atlassian.net/browse/DC-393

#### ✍️ Description

In CO, `page_load` events were appearing in the analytics data layer but never reaching Amplitude. Root cause is symmetric across states, but only surfaced in CO because the deployed environment has Amplitude configured without Mixpanel.

The analytics package emits two distinct DOM CustomEvents:
- `digitalData:PageViewed` — fired by `DataLayer.pageLoad()`
- `digitalData:EventTracked` — fired by `DataLayer.trackEvent()`

`mixpanel-bridge` listens to both. `amplitude-bridge` only listened to `EventTracked` (an explicit test even asserted "does not forward PageViewed events"). In DC, both Mixpanel and Amplitude tokens are configured, so page_loads flow to Mixpanel and the gap was invisible. In CO the Mixpanel token isn't set, the `MixpanelAnalytics` component doesn't render, and page_loads have nowhere to go.

**Fix:** Extend `amplitude-bridge` to also forward `PageViewed` events to `amplitude.track('page_load', eventData)`, mirroring the existing mixpanel-bridge pattern. Both listeners share a single `forward()` handler that pulls `eventName` and `eventData` from the CustomEvent detail.

**Test changes:**
- Removed the old negative assertion ("does not forward PageViewed events")
- Added a positive test asserting `pageLoad({ name: 'Dashboard' })` results in `amplitude.track('page_load', { name: 'Dashboard', ...page context })`
- Updated the file-level JSDoc to reflect that both event types are forwarded and to explain why `defaultTracking: false` is kept (page tracking is opt-in via the data layer rather than Amplitude's autocapture)

Analytics package suite: 92 passing.

#### 🔗 Links to related PRs
N/A (single-repo change)

#### ✅ Completion tasks

- [x] Added relevant tests
- [ ] Configuration changes:


adbergen:
# QA Summary — DC-393

## What Changed

Page-load analytics events now reach Amplitude in addition to Mixpanel. Previously, every page load in the CO portal was visible in the data layer (`window.digitalData`) but no corresponding event was sent to Amplitude. After this fix, navigating to any page should send a `page_load` event to Amplitude. **No user-visible UI changes** — this is observability/instrumentation only.

## Affected Areas

- All pages in the **CO portal** — every page-load now fires an Amplitude event
- All pages in the **DC portal** — same instrumentation (regression check; DC was previously sending page_load via Mixpanel only)
- Amplitude live data (Amplitude project for the SUN Bucks portal)

## Test Cases

### Happy Path

- [ ] **CO portal — Amplitude receives page_load on every navigation.**
  1. Open the CO portal in a browser with DevTools open (Network tab).
  2. Filter network requests for `amplitude` (the SDK sends to a hostname containing "amplitude").
  3. Sign in as a CO test persona and navigate through several pages (login, dashboard, profile, address, card replacement flow, etc.).
  4. **Expected:** A new Amplitude network request fires on **every** page transition. Inspect the payload — each request body should include an event with `event_type: "page_load"` and `event_properties` containing fields like `application: "sebt-portal"`, `entry_source: "portal"`, `flow`, `name`, `step`, etc.

- [ ] **CO portal — Amplitude Live Stream confirmation.**
  1. Open the Amplitude project's "User Look-Up" or "Live Stream" view for the CO portal.
  2. Have a tester navigate around the CO portal in a separate browser.
  3. **Expected:** `page_load` events appear in Amplitude's Live Stream within a few seconds of each page transition, matching the visited pages.

### Edge Cases

- [ ] **Page name and step properties are populated.**
  1. Navigate to a deep page in CO (e.g. card replacement → confirm).
  2. Inspect the Amplitude `page_load` payload for that page.
  3. **Expected:** `name` and `step` fields are populated (not blank or "undefined") and match the page you're on.

- [ ] **Both Amplitude and Mixpanel fire (if Mixpanel is configured for the environment).**
  1. With DevTools Network tab open, navigate to a new page.
  2. **Expected:** If both `NEXT_PUBLIC_MIXPANEL_TOKEN` and `NEXT_PUBLIC_AMPLITUDE_API_KEY` are configured, you should see **both** an Amplitude request and a Mixpanel request fire for the same page load.

- [ ] **CTA click events still reach Amplitude (regression).**
  1. Click any tracked button (e.g. "Request a replacement card", "Sign in", "Apply now").
  2. **Expected:** An Amplitude network request fires with the corresponding event type (e.g. `replacement_card_cta`), as it did before this change.

### Regression Checks

- [ ] DC portal — page_load events continue to reach Mixpanel (no change expected; DC behavior was already correct).
- [ ] DC portal — page_load events also reach Amplitude (new behavior; verify symmetry with CO).
- [ ] No JavaScript errors in the browser console on any page navigation.
- [ ] The data layer (`window.digitalData.event` in DevTools console) still contains the page_load entries it always has.

## Test Credentials

Use any seeded CO portal persona. All test users follow the pattern `sebt.co+{scenario}@codeforamerica.org` for CO and `sebt.dc+{scenario}@codeforamerica.org` for DC.

- CO co-loaded household — `sebt.co+co-loaded@codeforamerica.org` — OTP retrieval depends on environment (see Environment Notes)
- CO verified household — `sebt.co+verified@codeforamerica.org`
- CO active card — `sebt.co+co-active@codeforamerica.org`
- DC verified household — `sebt.dc+verified@codeforamerica.org` (for the regression check)

## Environment Notes

- Verify on environments where the Amplitude project is configured. In local dev, `NEXT_PUBLIC_AMPLITUDE_API_KEY` must be set in `.env.local` for the Amplitude SDK to load at all — if the key is blank, the Amplitude bridge silently no-ops and you will see no network requests.
- For local OTP retrieval, use Mailpit at the local Mailpit URL (the docker-compose stack ships it alongside MSSQL).
- For deployed environments, the GitHub repo variable `AMPLITUDE_API_KEY` must be set per environment for the SDK to ship in the built bundle.
- **No new env vars, secrets, or app-settings were added in this change** — nothing to update in Tofu, AWS Secret Manager, or AppConfig.
- The fix lives in the shared analytics package, so any future portal app (not just CO and DC) will inherit the same `page_load`-to-Amplitude wiring automatically.

mtodd-cfa (approved):

PR #306, opened 2026-05-12
DC-410 | fix(security): bump Next.js to 16.2.6 to patch CVE-2026-45109 (DC-410)

jamesmblair:
## Jira ticket
[DC-410](https://codeforamerica.atlassian.net/browse/DC-410)

## Description
Patches **CVE-2026-45109** — a High-severity Middleware / Proxy bypass in App Router applications via segment-prefetch routes. This is the *incomplete-fix follow-up* to CVE-2026-44575: the original fix didn't cover `middleware.ts` under Turbopack, leaving the bypass exploitable without credentials.

- Advisory: https://github.com/advisories/GHSA-26hh-7cqf-hhc6
- Affected: Next.js `>=16.0.0, <16.2.6` (we were on `16.2.3`)
- Patched: `16.2.6` (Vercel coordinated security release, 2026-05-07)

The bump touches both web apps and the workspace test peers so a single `next@16.2.6` resolves across the lockfile:

- `src/SEBT.Portal.Web` — `16.2.3` → `16.2.6`
- `src/SEBT.EnrollmentChecker.Web` — `16.2.3` → `16.2.6`
- `packages/analytics` (devDep test peer) — `16.2.3` → `16.2.6`
- `packages/design-system` (devDep test peer) — `16.2.3` → `16.2.6`

### Cooldown exemption
`pnpm-workspace.yaml` enforces `minimumReleaseAge: 7d`. `16.2.6` was published 5 days ago, so pnpm would normally refuse to resolve it until 2026-05-14. A new version-pinned `minimumReleaseAgeExclude` block exempts only the exact `16.2.6` sibling set (`next` + `@next/env` + 8 `@next/swc-*` platform binaries). The exemption is scoped to this single version and ages itself out on the next bump.

**Integrity verified before exempting:**
- Published via npm OIDC trusted publisher (GitHub Actions on `vercel/next.js`)
- SLSA v1 provenance attestation present on registry
- Git tag `v16.2.6` → commit `ee6e79b1`, signed by Tim Neutkens (Next.js lead)
- Matches Vercel's [public coordinated May 2026 security release](https://vercel.com/changelog/next-js-may-2026-security-release)
- No compromise chatter found for `next@16.2.6`. The TanStack Shai-Hulud campaign (2026-05-11) hit separate package families (`@tanstack/react-router`, `@tanstack/start-*`) and post-dates this release. Our lockfile only resolves `@tanstack/react-query`, which is in the confirmed-clean family.

[...]

PR #309, opened 2026-05-12
DC-404 | fix(e2e): tolerate cache-bust query string in Playwright route mocks

jamesmblair:
## Jira ticket
[DC-404](https://codeforamerica.atlassian.net/browse/DC-404). Discovered while working on #306  

## Description
Fix for the e2e regression introduced by #305. Since that PR was merged, every PR's \`Playwright E2E Tests (co)\` and \`(dc)\` checks have been failing with a 30s timeout in \`page.fill('[name="streetAddress1"]')\`.

### Root cause
#305 added a per-request \`?_=<uuid>\` cache-busting query string to every GET issued through \`apiFetch\` (src/SEBT.Portal.Web/src/api/client.ts) as edge-cache leak defense. Playwright \`page.route()\` glob patterns match against the **full URL including the query string**, so:

| Pattern | URL on the wire | Matches? |
|---|---|---|
| \`**/api/household/data\` | \`/api/household/data\` | yes |
| \`**/api/household/data\` | \`/api/household/data?_=abc-123\` | no |

All three mocked GET endpoints — \`/api/auth/status\`, \`/api/household/data\`, \`/api/features\` — started falling through unhandled. With no household data and no feature flags, the address form stays in its loading state and never renders the \`streetAddress1\` input. Hence the 30s timeout.

### Fix
Append \`*\` to the three GET globs in \`src/SEBT.Portal.Web/e2e/fixtures/api-routes.ts\`. Playwright glob \`*\` matches any characters except \`/\`, so it matches both the bare path and the path-plus-query form. POST/PUT routes (\`household/address\`, \`cards/replace\`, \`auth/refresh\`) are left alone — cache-busting only applies to GET, so their URLs never gain a query string.

Added a short comment above the routes explaining why GET globs have the trailing \`*\`, so future readers don't strip it back out.

## Related PRs
- #305 — the cache-busting PR that introduced the wire-format change
- #306 — Next.js CVE bump, blocked by this same e2e regression. Will rebase main back in once this merges.

## Completion checklist
- [x] All three affected GET globs updated
- [x] POST/PUT globs deliberately left unchanged (not cache-busted)
- [x] Explanatory comment added so the \`*\` suffix isn't accidentally removed
- [ ] Playwright E2E Tests (co) green on this PR
- [ ] Playwright E2E Tests (dc) green on this PR

[DC-404]: https://codeforamerica.atlassian.net/browse/DC-404?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


ShaynaCummings (approved):
today I learned!!

PR #311, opened 2026-05-12
DC-356 security: add packageSourceMapping and NuGetAudit

robdiciuccio:
#### 🔗 Jira ticket
https://codeforamerica.atlassian.net/browse/DC-356

#### ✍️ Description
Adds the following NuGet security measures:
- `packageSourceMapping` to prevent internal package squatting in the public repo
- `NuGetAudit` vulnerability scanning for direct and transitive dependencies

NOTE: this does not negate the need for dependency **lockfiles** to prevent unwanted packages from being installed, but this will be addressed in a separate PR.

#### ✅ Completion tasks

- [ ] Added relevant tests
- [ ] Configuration changes:

### DIFF
--- a/Directory.Build.props
+++ b/Directory.Build.props
@@ -1,0 +1,26 @@
+<Project>
+
+  <!--
+    Repository-wide MSBuild defaults. Applied to every .csproj below this directory
+    via MSBuild's automatic import of Directory.Build.props.
+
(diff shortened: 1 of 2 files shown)


jamesmblair (approved):
I dug into this a bit more and think this is good-to-go as is. Looking forward to figuring out how to incorporate lockfiles.

PR #313, opened 2026-05-13
DC-360 - fix CTA copy key strings showing

necampanini:
#### 🔗 Jira ticket
[DC-360](https://codeforamerica.atlassian.net/browse/DC-360)

#### ✍️ Description
For households with no enrolled children (cancelled or pending applications), the dashboard previously rendered an info-alert that bled an unresolved i18n key (`actionNavigationSelfServiceUnavailable`) as its text, an empty "Enrolled Children" section, and a "Check existing cards" CTA whose scroll anchor would become dead once the empty section was removed. This change suppresses the alert entirely, hides the Enrolled Children section when `summerEbtCases.length === 0`, and gates the "Check existing cards" CTA on a new `hasCases` prop passed down from `DashboardContent`. Applies to both DC and CO state builds and only affects empty-case render paths — households with active cases continue to render the full dashboard unchanged.

#### 🔗 Links to related PRs
- {none — frontend-only change in the portal repo}

#### ✅ Completion tasks
- [ ] Added relevant tests
- [ ] Configuration changes:

[DC-360]: https://codeforamerica.atlassian.net/browse/DC-360?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


src/SEBT.Portal.Web/src/features/household/components/ActionButtons/ActionButtons.test.tsx
```diff
@@ -125,9 +125,33 @@ describe('ActionButtons', () => {
...
+  it('hides Check existing cards CTA when hasCases is false', () => {
+    render(
+      <ActionButtons
+        allowedActions={allowAll}
+        hasCases={false}
+      />
+    )
+    expect(screen.queryByText('Check existing cards')).toBeNull()
+    // Other CTAs unaffected
+    expect(screen.getByText('Change my mailing address')).toBeInTheDocument()
```
jamesmblair: If there are no cases, shouldn't the only available CTA be check existing applications?

jamesmblair (changes requested):

src/SEBT.Portal.Web/src/features/household/components/DashboardContent/DashboardContent.tsx
ShaynaCummings: Comparing to the [Figma](https://www.figma.com/design/32kDQ73MSbUNdAsVi8fQzF/CO-11-3-25-%F0%9F%9F%A2-Enrollment-Checker---Self-service-Portal--WORKING-?node-id=6784-32784&m=dev) design - I think we want to hide the EbtEdgeSection here as well, if there are no enrolled children (since they won't have cards).

ShaynaCummings: Actually - would we hide both the EnrolledChildren and EbtEdgeSection components altogether here if there are no enrolled children? instead of having EnrolledChildren just return null

ShaynaCummings (changes requested):

ShaynaCummings (approved):

jamesmblair (approved):

PR #327, opened 2026-05-15
DC-460 Portal-side exact-match filter for CO enrollment check candidates

jamesmblair:
#### 🔗 Jira ticket
https://codeforamerica.atlassian.net/browse/DC-460

#### ✍️ Description

Two changes that work together:

**Filter** (`enrollment_check_requires_at_least_one_exact_matched_field`, default `true` for CO): drops a Match/PossibleMatch result unless the birth year matches *and* either the full DOB or full name (case-insensitive) also matches. Error/NonMatch pass through. If all candidates are dropped for a child, a synthetic NonMatch is returned so the response always has one result per submitted child.

**PII boundary** (always-on): the handler replaces `FirstName`/`LastName`/`DateOfBirth` in every result with the submitted values before returning to the API, so no CBMS-normalized data (e.g. `JANE DOE`) ever surfaces to the UI.

The CO connector ([PR #40](https://github.com/codeforamerica/sebt-self-service-portal-co-connector/pull/40)) is updated to return real CBMS values instead of echoing submitted values — this is what makes the filter non-trivial, and what requires the PII boundary.

#### 🔗 Links to related PRs

- CO connector: https://github.com/codeforamerica/sebt-self-service-portal-co-connector/pull/40

#### ✅ Completion tasks

- [x] Added relevant tests
- [x] Configuration changes:

> **Note:** CO connector PR must merge first. Feature flag `enrollment_check_requires_at_least_one_exact_matched_field: true` must be set in AWS AppConfig for CO dev + prod.

### DIFF
--- a/src/SEBT.EnrollmentChecker.Web/src/lib/env.ts
+++ b/src/SEBT.EnrollmentChecker.Web/src/lib/env.ts
@@ -1,6 +1,14 @@
 import { createEnv } from '@t3-oss/env-nextjs'
 import { z } from 'zod'
 
+// z.coerce.boolean() uses Boolean() which treats any non-empty string as true,
+// so "false" would coerce to true. This handles the common env var string values.
+const boolEnv = (defaultValue = false) =>
+  z.preprocess(
+    (val) => (val === 'false' || val === '0' ? false : Boolean(val)),
+    z.boolean().optional().default(defaultValue)
+  )
+
 export const env = createEnv({
   server: {
     NODE_ENV: z.enum(['development', 'test', 'production']).optional(),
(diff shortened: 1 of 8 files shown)


src/SEBT.EnrollmentChecker.Web/src/lib/env.ts
```diff
@@ -1,6 +1,14 @@
 import { createEnv } from '@t3-oss/env-nextjs'
 import { z } from 'zod'
 
+// z.coerce.boolean() uses Boolean() which treats any non-empty string as true,
+// so "false" would coerce to true. This handles the common env var string values.
+const boolEnv = (defaultValue = false) =>
```
jamesmblair, line 6: Fixes an issue where any *defined* value was being evaluated as truthy, and therefore, `true`.

ShaynaCummings (approved):

ShaynaCummings (approved):

ShaynaCummings (approved):
👍

PR #320, opened 2026-05-15
DC-448 fix: bypass AuthGuard for id-proofing off-boarding 

adbergen:
#### 🔗 Jira ticket
[DC-448](https://codeforamerica.atlassian.net/browse/DC-448) — P1: in CO, declining consent on step-up takes you to an error screen and then logs you out.

#### ✍️ Description
When a CO user declines consent on the PingOne / Socure step-up flow, PingOne redirects them back to `/callback?error=…`. The callback page correctly forwards to `/login/id-proofing/off-boarding?reason=oidcCallbackError`, where the off-boarding page already renders the "We're sorry, we aren't able to show your Summer EBT information" copy that matches the [Step Up Failure Figma](https://www.figma.com/) (title, body, Back + Contact us buttons — design verified against the current `stepUpFailure` namespace and `OffBoardingContent` component).

The bug is upstream of the page itself. `/login/id-proofing/*` is wrapped in `IdProofingLayout`, which gates every child route with `AuthGuard`. The OIDC round-trip to PingOne can lose the portal session (cross-site cookie age, SameSite refusal on the inbound redirect, IdP-side `prompt=login` + step-up declining the new auth event). When that happens, `AuthGuard` sees `isAuthenticated === false` on the off-boarding page and bounces the user to `/login` — which produces exactly the "logs you out + ends up on a 404-ish screen" experience the ticket describes.

Fix: the off-boarding route is the terminal failure screen for the step-up round-trip. It must remain reachable regardless of session state. This PR adds a path-based bypass in `IdProofingLayout` so off-boarding renders without `AuthGuard`. The page itself already null-safes `session?.isCoLoaded`, so it works with or without a session.

#### 🔗 Links to related PRs
None.

#### ✅ Completion tasks

- [x] Added relevant tests
- [x] Meets acceptance criteria in ticket
- [ ] Configuration changes:
  - [ ] If new environment variables are added, update in Tofu
  - [ ] If new environment secrets are added, update in Tofu and set in AWS Secret Manager
  - [ ] If you're adding an appsetting, add it to the requisite example file, and update it in the AWS AppConfig
  - [ ] If appsetttings are changed, update in AWS AppConfig

[DC-448]: https://codeforamerica.atlassian.net/browse/DC-448?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

PR #334, opened 2026-05-19
fix: scope FeatureFlagsProvider to authenticated layout (DC-448 follow-up)

adbergen:
#### 🔗 Jira ticket
[DC-448](https://codeforamerica.atlassian.net/browse/DC-448) — P1: in CO, declining consent on step-up takes you to an error screen and then logs you out.

#### ✍️ Description
Follow-up to PR #320. The `AuthGuard` bypass landed correctly but the POC retest surfaced a second mechanism kicking users off the off-boarding route.

**Why the previous fix was incomplete.** `FeatureFlagsProvider` lived in the root `layout.tsx`, so it mounted on every route — including public ones (`/callback`, `/login/*`, `/login/id-proofing/off-boarding`). On mount it fires `GET /api/features`. The OIDC round-trip leaves the user's JWT cookie stale; on the next request, JwtBearer's `OnTokenValidated` session-lifetime check calls `context.Fail()` and the response goes back as **401**. `apiFetch` has a global "401 means session is dead → `window.location.replace('/login')`" rule with one exception for `/auth/status` (the bootstrap probe). `/features` wasn't on that list, so the user was hard-redirected to `/login` after `router.replace` brought them to off-boarding — they never saw the failure screen.

**Fix.** Move `FeatureFlagsProvider` out of the root layout and into `(authenticated)/layout.tsx`, alongside `AuthGuard`, `TokenRefresher`, and `UserDataSync`. The provider's only public-route consumer was `BetaBanner`, which moved with it.

- Public routes (`/callback`, `/login`, off-boarding) **never call `/features`** → no 401 cascade, off-boarding renders.
- Authenticated routes still call `/features` exactly as today. `AuthGuard` runs above the provider, so a `/features` 401 in that tree genuinely means the session just died mid-render — the existing `apiFetch` redirect is correct in that case.
- `useFeatureFlag` already returns `false` when used outside the provider, so the rearrangement is fail-safe: any future stray render of a flag-gated component outside the authenticated tree degrades to "feature off" rather than throwing.

The alternative fix (adding `/features` to the `apiFetch` bootstrap-probe list) was discarded because it would have introduced a flash-of-hidden-UI window on dead-session dashboard loads — every flag temporarily reads `false`, which is safe for today's flag set but a hazard if a future flag is added that defaults to a "show sensitive thing" state.

**Trade-off worth flagging:** `BetaBanner` is now scoped to authenticated routes. The login page no longer shows it. The banner exists to announce in-beta features, which only matter once a user is signed in — product should confirm this is acceptable before promoting beyond dev.

#### 🔗 Links to related PRs
- #320 (the original AuthGuard bypass, also for DC-448)

#### ✅ Completion tasks

- [x] Added relevant tests — _existing 804-test suite still passes; provider move is structural and exercised end-to-end by route tests_
- [ ] Configuration changes:

[DC-448]: https://codeforamerica.atlassian.net/browse/DC-448?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/SEBT.Portal.Web/src/app/(authenticated)/layout.tsx
+++ b/src/SEBT.Portal.Web/src/app/(authenticated)/layout.tsx
@@ -1,5 +1,7 @@
+import { BetaBanner } from '@/components/BetaBanner'
 import { AuthGuard, TokenRefresher } from '@/features/auth'
 import { UserDataSync } from '@/hooks/useUserDataSync'
+import { FeatureFlagsProvider } from '@/providers'
 import type { ReactNode } from 'react'
 
 interface AuthenticatedLayoutProps {
@@ -9,9 +11,12 @@ interface AuthenticatedLayoutProps {
 export default function AuthenticatedLayout({ children }: AuthenticatedLayoutProps) {
   return (
     <AuthGuard>
-      <TokenRefresher />
-      <UserDataSync />
-      {children}
+      <FeatureFlagsProvider>
+        <TokenRefresher />
+        <UserDataSync />
+        <BetaBanner />
+        {children}
+      </FeatureFlagsProvider>
     </AuthGuard>
   )
 }
--- a/src/SEBT.Portal.Web/src/app/layout.tsx
+++ b/src/SEBT.Portal.Web/src/app/layout.tsx
@@ -1,5 +1,4 @@
 import { AmplitudeAnalytics } from '@/components/AmplitudeAnalytics'
-import { BetaBanner } from '@/components/BetaBanner'
 import { MixpanelAnalytics } from '@/components/MixpanelAnalytics'
 import { SiteImproveAnalytics } from '@/components/SiteImproveAnalytics'
 import { primaryFont } from '@/design/fonts'
@@ -8,7 +7,6 @@ import {
   AuthProvider,
   AxeProvider,
   DataLayerProvider,
-  FeatureFlagsProvider,
   I18nProvider,
   QueryProvider
 } from '@/providers'
@@ -113,23 +111,21 @@ export default async function RootLayout({
         >
           <QueryProvider>
             <AuthProvider>
-              <FeatureFlagsProvider>
-                <I18nProvider>
-                  <SkipNav />
-                  <AxeProvider>
-                    {/* Portal target for page-level alerts rendered above the header.
-                        Currently used by AddressForm (30-char street address error).
-                        If a second consumer appears, refactor to a SiteAlertContext so
-                        child components call setSiteAlert() instead of using createPortal directly. */}
-                    <div id="site-alerts" />
-                    <BetaBanner />
-                    <Header state={state} />
-                    <main id="main-content">{children}</main>
-                    <HelpSection state={state} />
-                    <Footer state={state} />
-                  </AxeProvider>
-                </I18nProvider>
-              </FeatureFlagsProvider>
+              <I18nProvider>
+                <SkipNav />
+                <AxeProvider>
+                  {/* Portal target for page-level alerts rendered above the header.
+                      Currently used by AddressForm (30-char street address error).
+                      If a second consumer appears, refactor to a SiteAlertContext so
+                      child components call setSiteAlert() instead of using createPortal directly. */}
+                  <div id="site-alerts" />
+                  {/* FeatureFlagsProvider lives in (authenticated) so public routes don't fire /features. */}
+                  <Header state={state} />
+                  <main id="main-content">{children}</main>
+                  <HelpSection state={state} />
+                  <Footer state={state} />
+                </AxeProvider>
+              </I18nProvider>
             </AuthProvider>
           </QueryProvider>
         </DataLayerProvider>


adbergen:
# QA Summary — DC-448 (follow-up)

## What Changed

This is a follow-up to the earlier DC-448 fix (PR #320). POC testing showed the original fix was incomplete — declining consent on the CO step-up flow was still bouncing users to the login page instead of the "We're sorry, we aren't able to show your Summer EBT information" screen. This PR removes the remaining cause: a feature-flag fetch the portal was making on every page (including public pages) that was failing with an authentication error after the OIDC round-trip, which a global rule was treating as "session dead, redirect to login."

After this PR, the feature-flag setup only runs on signed-in pages. Declining consent on Socure now reliably lands users on the step-up failure screen.

**One intentional side effect:** the small "site is in beta" notice banner (the thin info alert at the top of the page) is now only visible to signed-in users. It no longer appears on the login page or the step-up failure screen.

## Affected Areas

- CO Portal — Step-up identity-verification decline flow (the specific bug DC-448 tracks).
- CO Portal and DC Portal — The "site in beta" notice banner: now only appears once a user is signed in. Public/pre-sign-in pages no longer show it.
- CO Portal and DC Portal — Public/unauthenticated pages (login, the step-up failure screen, the OIDC callback page) should load slightly faster — one fewer API call on each.

## Test Cases

### Happy Path

- [ ] **Declining Socure consent now lands on the step-up failure screen, not the login page.**
  1. Sign in to the CO Portal as a user who has not yet completed step-up identity verification.
  2. Navigate to any page that triggers the step-up gate (the dashboard's Verify prompt will work).
  3. Click **Verify**. You'll be redirected to MyColorado / PingOne and eventually to the Socure consent prompt.
  4. On the Socure consent prompt, click **Decline** / refuse consent.
  5. **Expected:** The browser returns to the portal and renders the step-up failure screen. Title: "We're sorry, we aren't able to show your Summer EBT information." Body: "You can contact us if you need more help." Two buttons visible: **Back** and **Contact us**.
  6. **Expected:** The user is NOT redirected to the sign-in page. The page renders without a flash or visible redirect.

- [ ] **Beta banner still appears on every signed-in page.**
  1. Sign in to either portal (CO or DC).
  2. Navigate through several signed-in pages (dashboard, cards info, profile/address, etc.).
  3. **Expected:** If the beta banner is enabled in this environment, it appears at the top of every signed-in page exactly as before this PR.

### Edge Cases

- [ ] **The step-up failure screen renders even if the session is fully gone.**
  1. After landing on the step-up failure screen (via the decline flow above), open DevTools → Application → Cookies and manually delete the portal session cookies.
  2. Refresh the page.
  3. **Expected:** The step-up failure screen still renders. No redirect to login.

- [ ] **The beta banner no longer appears on public/pre-sign-in pages.**
  1. Sign out completely. Visit the login page.
  2. **Expected:** No beta banner is visible at the top of the login page (even if the feature flag is enabled).
  3. Sign in.
  4. **Expected:** The beta banner now appears once you're authenticated.

### Regression Checks

- [ ] **Sign-in and sign-out flows behave as before.**
  1. Sign in to each portal (CO and DC) using the existing methods (MyColorado for CO, OTP for DC).
  2. Sign out from the dashboard.
  3. Sign in again.
  4. **Expected:** No new errors, no broken redirects. Normal flow.

- [ ] **Successful step-up still completes end-to-end (CO).**
  1. Sign in to the CO Portal. Click **Verify** to start step-up.
  2. Approve consent on Socure and complete the identity-proofing flow to a successful outcome.
  3. **Expected:** User lands on the dashboard with elevated verification. No regression in the success path.

- [ ] **Dead-session redirect still works.**
  1. Sign in to the DC Portal. From the dashboard, open DevTools → Application → Cookies and delete the portal session cookies.
  2. Navigate to any other authenticated page (cards info, profile/address, etc.).
  3. **Expected:** User is redirected to the sign-in page — the existing session-protection behavior is preserved.

- [ ] **Other authenticated pages render correctly.**
  1. Sign in to either portal. Visit dashboard, cards info, replacement card flows, address-update flow.
  2. **Expected:** All pages render normally. Feature-flag-gated UI (case numbers, last-4 card digits, contact preferences) appears/hides exactly as before this PR.

## Test Credentials

- **DC portal** — use any seeded DC persona email of the form `sebt.dc+<scenario>@codeforamerica.org` (e.g. `sebt.dc+verified@codeforamerica.org`). OTP is delivered to the seeded inbox in the lower environment.
- **CO portal** — sign-in is via MyColorado (PingOne). Use the existing MyColorado test account for the lower-environment PingOne tenant. Coordinate with the CO state-connector team if you don't have credentials.

## Environment Notes

- The CO step-up decline flow must be tested against a deployed environment that integrates with the real PingOne + Socure. Mock data bypasses OIDC so the bug can't surface locally.
- After this PR merges, confirm the deployed image was rebuilt from this commit (or later) before testing.
- Product should review the beta-banner side effect (banner no longer shown on public pages) and confirm it's acceptable. If the banner needs to remain on the login page, this can be revisited separately.
- No env-var, secret, AppConfig, or Tofu changes are needed.

PR #339, opened 2026-05-20
feat: add Remove link to review screen (DC-336)

adbergen:
#### 🔗 Jira ticket
[DC-336](https://codeforamerica.atlassian.net/browse/DC-336) — P2: Add links to remove children from Review screen in enrollment checker.

#### ✍️ Description
The Enrollment Checker review page already let users update a child's info but had no way to drop one — meaning info typed into the form would always end up in the CBMS check request, even if the user changed their mind. This PR adds a Remove link next to "Update this child's information" on each child card, matching the Figma (red error-color link, sits to the right of the update link).

**Wiring:**
- `removeChild` already existed on `EnrollmentContext`; this PR just exposes it through `ReviewPage` → `ChildReviewCard` as a new `onRemove` prop.
- Copy comes from the new `GLOBAL - Remove` content row ("Remove" / "Eliminar"), resolved via `tCommon('remove')`.
- Red styling via the USWDS `text-error-dark` utility class — no custom CSS.

**Drive-by test fix:** `ReviewPage.test.tsx`'s `Seeder` helper used to call `addChild` directly in the render body, which re-fired on every parent re-render and quietly seeded duplicate children. Existing tests didn't catch it because they query for unique buttons like "Add another child". The new Remove test exposed it (multiple per-child Remove buttons matched a single `getByRole` query). Fixed by moving the seed into `useEffect` with empty deps, plus a `beforeEach` that clears `sessionStorage` to keep state from bleeding between tests.

#### 🔗 Links to related PRs
None.

#### ✅ Completion tasks

- [x] Added relevant tests
- [ ] Configuration changes:

[DC-336]: https://codeforamerica.atlassian.net/browse/DC-336?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


adbergen:
# QA Summary — DC-336

## What Changed

The CO Enrollment Checker's review screen now has a "Remove" link on every child row, sitting next to the existing "Update this child's information" link. Before this PR, a user could only update a child's details — never remove the child entirely. After clicking Remove, the child's info is dropped from the list and is NOT included in the eligibility-check request when the user clicks Submit.

Spanish copy: **Eliminar**.

## Affected Areas

- CO Enrollment Checker — Review screen (the page that appears after the user has added at least one child and before they click Submit).

## Test Cases

### Happy Path

- [ ] **The Remove link appears on every child row.**
  1. Open the CO Enrollment Checker landing page and click through the disclaimer.
  2. Add one child via the child form and continue to the review screen.
  3. **Expected:** The child's name and birthdate appear, with two links underneath: "Update this child's information" (teal) and "Remove" (red).

- [ ] **Clicking Remove drops that child from the review list.**
  1. From the review screen with one child listed, click **Remove**.
  2. **Expected:** The child immediately disappears from the list. No confirmation dialog is shown — the removal is instant.

- [ ] **Submitting after removal does NOT include the removed child.**
  1. Add two children — note the first child's name (call it Child A) and the second's (Child B).
  2. On the review screen, click **Remove** on Child A.
  3. Click **Submit** and continue through to the results page.
  4. **Expected:** The results page shows ONLY Child B. There is no row for Child A.

- [ ] **Remove works on the right row in a multi-child list.**
  1. Add three children (A, B, C).
  2. On the review screen, click **Remove** on the middle child (B).
  3. **Expected:** B disappears. A and C remain, in their original order.

### Edge Cases

- [ ] **Removing the last child empties the review list.**
  1. Add a single child and reach the review screen.
  2. Click **Remove**.
  3. **Expected:** The list is now empty. Submit can be clicked, but with no children to check, the result is whatever the backend returns for an empty submission. (Confirm this matches existing/expected behavior — out of scope for this PR.)

- [ ] **After removing, adding another child works normally.**
  1. Add a child, reach the review screen, click **Remove**, then click **Add another child**.
  2. Fill out the form for a new child and continue to review.
  3. **Expected:** The new child appears on the review screen exactly as if no prior child had been added.

- [ ] **After removing, "Update this child's information" still works for remaining children.**
  1. Add two children. Remove the first. Click **Update this child's information** on the remaining child.
  2. Edit the child's details and continue back to review.
  3. **Expected:** The updated child appears with the new details. No regression to the existing update flow.

### Regression Checks

- [ ] **Existing review-screen behavior is unchanged for users who don't click Remove.**
  1. Add a child, reach the review screen, click **Submit** WITHOUT clicking Remove.
  2. **Expected:** Results page renders as before this PR. No new errors, no missing UI.

- [ ] **Spanish copy reads "Eliminar".**
  1. Switch the language toggle to Spanish.
  2. Reach the review screen with at least one child.
  3. **Expected:** The Remove link reads **Eliminar**.

- [ ] **"Update this child's information" link still works.**
  1. On the review screen with at least one child, click **Update this child's information**.
  2. **Expected:** User is taken back to the child form with that child's details pre-filled — same as before this PR.

## Test Credentials

Not applicable — the CO Enrollment Checker is a public, unauthenticated tool. No sign-in is required.

## Environment Notes

- Test on the CO Enrollment Checker app (not the DC portal). The Remove link is wired into the CO checker review screen.
- No env-var, secret, AppConfig, or Tofu changes.
- The "Remove" copy was added to the content spreadsheet and the CSV in this PR — no separate content-team coordination needed.

src/SEBT.EnrollmentChecker.Web/src/features/enrollment/components/ChildReviewCard.tsx
```diff
@@ -31,13 +33,22 @@ export function ChildReviewCard({ child, onEdit }: ChildReviewCardProps) {
...
-      <button
-        type="button"
-        className="usa-link usa-button--unstyled"
-        onClick={() => onEdit(child.id)}
-      >
-        {t('tableAction')}
-      </button>
+      <div className="display-flex flex-row flex-wrap gap-2">
+        <button
+          type="button"
```
necampanini, line 38: might be worth adding labels per button for screen reader accessibility. as is, I think a screen reader will just read out "remove, remove, remove". 

we could give each action button a per-child accessible name, `aria-label={${tCommon('remove')} ${fullName}}`. the Edit button could use something like this too to be consistent

src/SEBT.EnrollmentChecker.Web/src/features/enrollment/components/ReviewPage.tsx
necampanini, line 25: do we handle an empty state if all children are removed? would the `Submit` button still be active?

necampanini (changes requested):
looks good, just small accessibility changes requested

adbergen:
Addressed both review items in 53fc4092:

- **Accessibility:** added per-child aria-labels to both the Edit and Remove buttons, of the form `"${actionLabel}: ${fullName}"`. Screen readers now read "Update this child's information: Jane Doe" / "Remove: Jane Doe" instead of repeating the bare action text across rows. New test asserts both buttons carry the child name in their accessible name.
- **Empty state:** Submit is now `disabled` when `state.children.length === 0`. A user who removes their last child can't accidentally submit an empty check. New test covers the disabled state + click-is-noop in one shot.

necampanini (approved):

PR #343, opened 2026-05-21
DC-148 / DC-473 analytics bridges & enrollment checker wiring

mtodd-cfa:
---
[DC-148][DC-473] Extract analytics bridge components to shared package and wire up enrollment checker
---

#### 🔗 Jira ticket

- https://codeforamerica.atlassian.net/browse/DC-473
- https://codeforamerica.atlassian.net/browse/DC-148

#### ✍️ Description
Moves the three vendor analytics bridge components (AmplitudeAnalytics, MixpanelAnalytics, SiteImproveAnalytics) from the portal's local src/components/ into the shared packages/analytics package so the
enrollment checker can use them without duplication. Wires all three analytics vendors into the enrollment checker, and threads the required env vars through env.ts, both Dockerfiles, and the CI workflow.

packages/analytics changes:
- Added AmplitudeAnalytics, MixpanelAnalytics, and SiteImproveAnalytics components with tests; exported from index.ts
- AmplitudeAnalytics accepts amplitude as a prop (AmplitudeLike) instead of importing @amplitude/analytics-browser directly — keeps the package free of runtime dependencies; callers supply their own import
- SiteImproveAnalytics renders unconditionally (no state prop); callers gate on the env var being present, consistent with how Amplitude and Mixpanel work
- Split Mixpanel boot snippet from CDN load to fix onReady and init errors

Portal changes:
- Deleted local AmplitudeAnalytics, MixpanelAnalytics, SiteImproveAnalytics components and their tests (superseded by the shared package)
- Updated layout.tsx to import from @sebt/analytics and pass the amplitude instance as a prop

Enrollment checker changes:
- Wired all four vendors (GA, Amplitude, Mixpanel, SiteImprove) in layout.tsx
- Added NEXT_PUBLIC_GA_ID, NEXT_PUBLIC_AMPLITUDE_API_KEY, NEXT_PUBLIC_MIXPANEL_TOKEN, NEXT_PUBLIC_SITEIMPROVE_ID to env.ts, both Dockerfiles, and deploy-enrollment-checker.yaml

Test fix:
- Added in-memory localStorage polyfill to test-env-preload.ts to fix 4 pre-existing test failures in Node.js 25 (native localStorage stub has no methods when --localstorage-file is not configured)

#### 🔗 Links to related PRs

N/A

#### ✅ Completion tasks
- [x] Added relevant tests
- [ ] Configuration changes:

[DC-148]: https://codeforamerica.atlassian.net/browse/DC-148?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


src/SEBT.Portal.Web/src/app/layout.tsx
```diff
@@ -159,9 +157,8 @@ export default async function RootLayout({
       )}
       {/* Amplitude - only rendered when NEXT_PUBLIC_AMPLITUDE_API_KEY is configured */}
       {amplitudeApiKey && <AmplitudeAnalytics apiKey={amplitudeApiKey} />}
-      {/* SiteImprove — DC-only per DC-272. Gated by both state and env var so an
-          accidentally-set NEXT_PUBLIC_SITEIMPROVE_ID in another state does not enable it. */}
-      {state === 'dc' && siteImproveId && (
+      {/* SiteImprove — only rendered when NEXT_PUBLIC_SITEIMPROVE_ID is configured */}
```
mtodd-cfa, line 160: Noting that this change removed the hardcoded "DC" specific guard

adbergen:
### Code review

Found 1 issue:

1. Mixpanel CDN host `cdn.mxpnl.com` is missing from `script-src` in `proxy.ts`. The PR splits the Mixpanel boot so the SDK is now loaded as a declarative `<Script src="https://cdn.mxpnl.com/libs/mixpanel-2-latest.min.js">` tag in [`packages/analytics/src/MixpanelAnalytics.tsx`](https://github.com/codeforamerica/sebt-self-service-portal/blob/2bc3f1b00edcb483a951b53a1c296bc6fb41a0b5/packages/analytics/src/MixpanelAnalytics.tsx#L40-L45) rather than being injected by an inline boot snippet (which previously inherited trust via the nonce + `'strict-dynamic'`). `'strict-dynamic'` only propagates trust to scripts *programmatically* injected by an already-trusted script — it does not exempt declarative `<script src=...>` tags, so production CSP will block the CDN load. `connect-src` correctly gets `https://api-js.mixpanel.com` (gated by `hasMixpanel`) but the matching `script-src` entry was not added.

CLAUDE.md: "When adding any browser-side call to a new external domain (API, SDK, analytics, fonts), add the domain to the appropriate CSP directive… A missing entry means the feature silently fails in production."

https://github.com/codeforamerica/sebt-self-service-portal/blob/2bc3f1b00edcb483a951b53a1c296bc6fb41a0b5/src/SEBT.Portal.Web/src/proxy.ts#L68-L78

🤖 Generated with [Claude Code](https://claude.ai/code)

<sub>- If this code review was useful, please react with 👍. Otherwise, react with 👎.</sub>

adbergen (changes requested):
Nice work. Just one change requested.

adbergen (approved):
LGTM!

PR #366, opened 2026-05-29
DC-374(Bug): check authentication before allowing users to go to Callback

noxmwalsh:
#### 🔗 Jira ticket

[DC-374](https://codeforamerica.atlassian.net/browse/DC-374)

#### ✍️ Description

After a successful OIDC login, using the browser back button from the dashboard could land users on `/callback` again (or on a stale callback URL without query params). The callback page treated that as a failed login and sent them to id-proofing off-boarding with `reason=oidcCallbackError` ("We aren't able to show your Summer EBT information").

This change makes `/callback` a **one-time handoff** for unauthenticated users only:

- Wait until `/auth/status` has finished loading before treating missing `code` or `state` as an error.
- If the visitor already has a portal session (`isAuthenticated`), redirect to `/dashboard` and skip the OIDC token exchange, including when the URL still has a used authorization code or no query string at all.
- Unauthenticated visitors keep the existing behavior: IdP errors, missing params, and exchange failures still route to off-boarding.

Effect deps now use `isAuthenticated` and `isLoading` instead of `session`, so a session update after `login()` does not re-run the effect and cancel an in-flight exchange.

#### 🔗 Links to related PRs

None

#### ✅ Completion tasks

- [x] Added relevant tests
- [ ] Configuration changes: **N/A**
  - [ ] If appsettings are changed, update in AWS AppConfig

#### Test plan

1. Log in through MyCO/OIDC until you reach the dashboard.
2. Press the browser back button.
3. Confirm you return to `/dashboard` (or stay in the app) and do not see the off-boarding "can't show your information" screen.
4. Repeat with back from dashboard when history includes `/callback` with no query params and with a stale `?code=` and `?state=` in the URL.
5. Confirm a logged-out visit to `/callback` without params still goes to off-boarding.
6. Run `cd src/SEBT.Portal.Web && pnpm test -- src/app/(public)/callback/page.test.tsx --run`.

[DC-374]: https://codeforamerica.atlassian.net/browse/DC-374?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/SEBT.Portal.Web/src/app/(public)/callback/page.tsx
+++ b/src/SEBT.Portal.Web/src/app/(public)/callback/page.tsx
@@ -8,9 +8,10 @@ import {
   OidcCallbackTokenResponseSchema,
   OidcCompleteLoginResponseSchema
 } from '@/features/auth/api/oidc'
+import { hasIal1Plus, isIdProofingCompletionFresh } from '@/lib/jwt'
 import { getState } from '@sebt/design-system'
 import { useRouter } from 'next/navigation'
-import { useEffect, useRef } from 'react'
+import { useEffect, useMemo, useRef } from 'react'
 import { useTranslation } from 'react-i18next'
 
 /**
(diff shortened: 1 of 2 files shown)


src/SEBT.Portal.Web/src/app/(public)/callback/page.tsx
```diff
@@ -26,16 +25,28 @@ import { useTranslation } from 'react-i18next'
...
   const exchangeStartedRef = useRef(false)
   const isCO = getState() === 'co'
 
   useEffect(() => {
+    if (isLoading) {
+      return
+    }
+
+    if (isAuthenticated) {
+      router.replace('/dashboard')
```
necampanini, line 68: could be wrong but won't this skip over a users in-progress proofing?   

this previous line handled that ` if (hasIal1Plus(session) && isIdProofingCompletionFresh(session))`

I think we can have a `const isProofedReEntry = hasIal1Plus(session) && isIdProofingCompletionFresh(session)` outside of the effect, and add `isProofedReEntry` added to the dependency list, we'd handle it as well

necampanini (changes requested):
looks good, but I'm unsure about the short circuiting with the `isAuthenticated`

src/SEBT.Portal.Web/src/app/(public)/callback/page.tsx
noxmwalsh: Ack!  That's actually critical to the step-up process, so we do want to handle this as suggested. I went ahead and did that in e0ab7f03b05388537a5463533281b9859da161a8

I also addressed another issue for handling IdP errors that come **AFTER** being logged in. see: e1d12950ece2114f0317ebb3e549965cc92b6906

necampanini (approved):

PR #379, opened 2026-06-04
DC-457(Bug): Fix stale mailing address when navigating back

noxmwalsh:
#### 🔗 Jira ticket
https://codeforamerica.atlassian.net/browse/DC-457

#### ✍️ Description
In DC, users who updated their mailing address and then clicked Back could see the old address on the dashboard or profile, or on the address form after returning from the suggested-address step.

The replacement-card step showed the new address from address-flow context, but the dashboard and profile read `addressOnFile` from the React Query `householdData` cache. A successful update only invalidated that query, so the UI could keep showing stale data until refetch finished, or overwrite an optimistic update if the refetch returned old data.

This change patches `addressOnFile` in the cache as soon as a PUT returns `status: valid`, using `normalizedAddress` from the response or the submitted address when normalization is omitted. Background invalidation still runs for consistency with the server.

When the user goes Back from the suggested-address or address-not-found screens, validation outcome is cleared without discarding `enteredAddress`, and the address form seeds fields from that value so edits are not replaced by `addressOnFile`.

E2E route mocks now update `household/data` after a successful address PUT so tests match production behavior. Unit and E2E regression tests cover both flows.

#### 🔗 Links to related PRs
None

#### ✅ Completion tasks

- [x] Added relevant tests
- [ ] Configuration changes:

## Test plan

- [ ] DC: From dashboard or profile, open Change mailing address, submit a new valid address, continue to replacement cards, click Back. Confirm the dashboard mailing address matches what was submitted.
- [ ] DC: Submit an address that routes to suggested-address, click Back. Confirm the form still shows what you typed, not the previous address on file.
- [ ] Regression: `pnpm exec vitest run src/features/address/api/client.test.ts` in `src/SEBT.Portal.Web`
- [ ] Regression: `NEXT_PUBLIC_STATE=dc pnpm exec playwright test e2e/card-replacement/address-back-stale-cache.spec.ts e2e/card-replacement/address-form-back-suggestion.spec.ts e2e/card-replacement/address-flow.spec.ts --project=chromium` in `src/SEBT.Portal.Web`


src/SEBT.Portal.Web/src/features/address/context/AddressFlowContext.tsx
```diff
@@ -21,8 +21,12 @@ interface AddressFlowContextValue {
   enteredAddress: UpdateAddressRequest | null
   /** Store the validation result and the entered address together after backend response. */
   setValidationResult: (result: AddressUpdateResponse, entered: UpdateAddressRequest) => void
-  /** Clear validation state (e.g. when navigating back to the address form). */
+  /** Clear validation outcome only; keeps enteredAddress for form re-edit. */
+  clearValidationOutcome: () => void
```
spokenbird, line 25: Nit: the names ClearValidationOutcome and ClearValidationResult are very similar even though the method doc calls out the difference, maybe more clear names could help clarify intent?

Something like clearValidationKeepEntered and clearValidationAndEntered? just a thought.

spokenbird (approved):
Looks good with a small non-blocking nit

PR #380, opened 2026-06-04
DC-343 - Hot-reload AppConfig via host-owned background service

spokenbird:
#### 🔗 Jira ticket
DC-343

#### ✍️  Description

  **Problem:** AppConfig values changed in AWS were not picked up by the running ECS
  task — the only way to surface a new value was a redeploy.
  
  **Root cause (differs from the ticket's original hypothesis).** The ticket suspected a
  disposed `LoggerFactory` stopping the reload timer; that was disproven. The real cause is
  two compounding issues:

  1. **No reload loop.** The provider loaded once at startup and nothing polled afterward.
  2. **The change-notification bridge is severed at startup.** During plugin registration
     (`AddPlugins`), a transient `ServiceProvider` is built to eagerly construct health-check
     plugins. Because the host registers `IConfiguration` (the `ConfigurationManager`) via a
     factory, that transient provider takes ownership of it and **disposes it** when it's torn
     down. `ConfigurationManager.Dispose()` disposes the provider→root change-token
     registrations (and the providers themselves) but leaves them in the providers list. Net
     effect: configuration *reads* keep working, but a provider reload's `OnReload()` no longer
     reaches the configuration root's change token — so `IFeatureManager` and every
     `IOptionsMonitor<T>` consumer keep serving stale values until the next restart.

  **Fix.**  
  - A host-owned `AppConfigAgentReloadService` (`BackgroundService` + `PeriodicTimer`) polls
    each AppConfig provider on an interval (default 90s).
  - The provider survives the spurious disposal — it no longer disposes its `SemaphoreSlim`, so
    reloads can still acquire the lock and refresh `IConfiguration`.
  - On a **real** change, the reload service calls `IConfigurationRoot.Reload()`, which raises
    the root change token directly — bypassing the severed bridge — so all cached consumers
    refresh even though the `ConfigurationManager` was disposed. Reloads are change-gated, so
    unchanged polls don't churn consumers.
  
  **Why not fix the disposal at the source?** A transient `ServiceProvider` disposing the host
  `ConfigurationManager` is a known ASP.NET Core footgun and could recur from any library that
  builds and disposes a child provider. Re-firing the root token from the component we own is
  robust to all such cases. Hardening `AddPlugins` is noted as separate follow-up tech debt.
  
  **Testing.**
  - Host-level tests reproduce the bug on a developer machine and verify the fix
    end-to-end: `IFeatureManager` refreshes after a reload through the disposed manager, and
    `IConfigurationRoot.Reload()` preserves runtime-set values such as the connection string.
  - Validated on dev-dc with no redeploy: a feature-flag change (`enable_beta_banner`) hot-reloaded,
    and an `IOptionsMonitor` change (`SelfServiceRules:AddressUpdate:Enabled`) flipped a household's
    available actions — covering both the `IFeatureManager` and `IOptionsMonitor<T>` consumer paths.
  - Same tests as above ran and confirmed on CO

  #### 🔗 Links to related PRs  
  None.

  #### ✅ Completion tasks
  
  - [x] Added relevant tests
  - [x] Configuration changes:
    - [x] If new environment variables are added, update in Tofu — N/A (none added)
    - [x] If new environment secrets are added, update in Tofu and set in AWS Secret Manager — N/A (none added)
    - [x] If you're adding an appsetting, add it to the requisite example file, and update it in the AWS AppConfig — N/A (no required appsetting; `AppConfig:Agent:ReloadAfterSeconds` is optional, defaults to 90)
    - [x] If appsettings are changed, update in AWS AppConfig — N/A


src/SEBT.Portal.Infrastructure/Configuration/AppConfigAgentConfigurationProvider.cs
```diff
@@ -332,7 +365,7 @@ public void Dispose()
     public override string ToString()
     {
         var className = GetType().Name;
-        var profile = $"{_profile.ApplicationId}:{_profile.EnvironmentId}:{_profile.ProfileId}:{_profile.ReloadAfterSeconds}";
+        var profile = $"{_profile.ApplicationId}:{_profile.EnvironmentId}:{_profile.ProfileId}";
         var isFeatureFlag = _profile.IsFeatureFlag ? " (Feature Flag)" : string.Empty;
 
         return $"{className} - {profile}{isFeatureFlag}";
```
noxmwalsh, line 371: We skip disposing `SemaphoreSlim` so hot-reload survives host disposal, but still dispose the owned HttpClient here. Production uses `ownsHttpClient: true`, so if Dispose() runs during the AddPlugins startup teardown, could that break later ReloadAsync() calls?

test/SEBT.Portal.Tests/Unit/Configuration/AppConfigAgentConfigurationProviderTests.cs
```diff
@@ -1037,6 +956,130 @@ public void Load_SubsequentReloads_ShouldNotRetryOnFailure()
...
+        await provider.ReloadAsync(); // agent returns the same data again
+
+        // Assert
+        Assert.False(token.HasChanged); // unchanged config must NOT raise a reload
+        Assert.True(provider.TryGet("FeatureManagement:feature1", out var value));
+        Assert.Equal("true", value);
+    }
+
+    [Fact]
+    public async Task ReloadAsync_AfterDispose_StillReloads()
```
noxmwalsh, line 992: If https://github.com/codeforamerica/sebt-self-service-portal/pull/380/changes#r3375900060 sounds like a concern, since we're covering and proved the non owned-httpclient path?

For example, `ReloadAsync_AfterDispose_StillReloads` uses `ownsHttpClient: false`, so it doesn’t cover the production path.

noxmwalsh (changes requested):

src/SEBT.Portal.Infrastructure/Configuration/AppConfigAgentConfigurationProvider.cs
spokenbird: Good catch! 

For the production path, I believe ownsHttpClient is only true when no HttpClient is supplied to the source (Build() does var ownsHttpClient = HttpClient == null) but AddAppConfigAgent always passes HttpClient so I believe ownsHttpClient was false and Dispose() never touched the client. I think that's why we didn't see the issue you point out when I tested in AWS. 

That said, it could still be a potential issue in the future if we ever construct the provider without an owned client. For that reason, I removed the owned client entirely and updated related tests.

noxmwalsh (approved):

PR #405, opened 2026-06-16
DC-545: Treat .NET compiler warnings as errors

jamesmblair:
#### 🔗 Jira ticket
https://codeforamerica.atlassian.net/browse/DC-545

#### ✍️ Description
Sets `TreatWarningsAsErrors=true` in `Directory.Build.props`, applying it solution-wide via MSBuild's automatic import. The build was already at zero warnings before this change; the preceding commits clear the pre-existing issues the flag surfaced:

- **Stale `IdProofingExpiresAt` writes** — 3 null-assignments in `DatabaseSeeder.cs` and a read in `UserEncryptedFieldMapper.cs` that DC-498 missed (CS0618)
- **Missing XML param tags** on `HouseholdController.GetHouseholdData` for parameters added since the doc comment was written (CS1573)
- **Missing XML doc comments** on Serilog enricher types — resolved by marking `PortalUserEnricher`, `OtelTracingSpanIdEnricher`, and their extension classes `internal`, which is the correct visibility for application-only plumbing (CS1591)
- **Obsolete `RedisBuilder()` constructor** in a test fixture, matching the `MsSqlBuilder` fix from DC-481 (CS0618)

Also adds a short README section documenting the policy and the `WarningsNotAsErrors` escape hatch for cases where a specific warning code needs to be demoted rather than fixed.

#### 🔗 Links to related PRs
- #406 — Retire `Users.IdProofingExpiresAt` column (follow-on, stacked on this PR)

#### ✅ Completion tasks

- [ ] Added relevant tests
- [x] Meets acceptance criteria in ticket
- [ ] Configuration changes: none

PR #406, opened 2026-06-16
Retire Users.IdProofingExpiresAt column

jamesmblair:
#### 🔗 Jira ticket
<\!-- No dedicated ticket — surfaced during DC-545 warnings-as-errors work -->

#### ✍️ Description
Stacked on #405.

The `Users.IdProofingExpiresAt` column was write-only: all three token-minting paths in `JwtTokenService` compute the `id_proofing_expires_at` JWT claim at runtime from `IdProofingCompletedAt + IdProofingValiditySettings.ValidityDays` and never read the stored value. Storing a derived, config-dependent expiration also meant any change to `ValidityDays` would require a bulk data update to existing rows.

This was surfaced during the warnings-as-errors work in #405: the `[Obsolete]` attribute on `User.IdProofingExpiresAt` flagged the remaining read in `UserEncryptedFieldMapper`, which led to confirming the column is fully redundant.

**Removed:**
- `User.IdProofingExpiresAt` domain model property (was `[Obsolete]`)
- `UserEntity.IdProofingExpiresAt` and its EF mapping
- All write paths: `DatabaseUserRepository` (update + create), `UserFactory`, `DataSeeder`
- `IdProofingExpiration` utility class (no remaining callers)
- Test assertions that verified the write-side computation and EF↔domain round-trip

**Added:**
- Migration to `DROP COLUMN Users.IdProofingExpiresAt`

#### 🔗 Links to related PRs
- #405 — Enable TreatWarningsAsErrors (base branch)

#### ✅ Completion tasks

- [ ] Configuration changes: none


src/SEBT.Portal.Infrastructure/Repositories/DatabaseUserRepository.cs
```diff
@@ -336,13 +335,6 @@ public async Task UpdateUserAsync(User user, CancellationToken cancellationToken
 
     private static string? NormalizeEmail(string? email) => EmailNormalizer.NormalizeOrNull(email);
 
-    /// <summary>
-    /// Persists a derived expiration for legacy DB readers. Runtime/JWT use
-    /// <see cref="IdProofingCompletedAt"/> + <see cref="IdProofingValiditySettings.ValidityDays"/>.
-    /// </summary>
-    private DateTime? ComputeStoredExpiration(DateTime? completedAt) =>
-        IdProofingExpiration.ComputeStoredExpiration(completedAt, _validitySettings.ValidityDays);
```
noxmwalsh, line 344: At line 25-ish:
```
    IOptions<IdProofingValiditySettings> validitySettings) : IUserRepository
{
    private readonly IdProofingValiditySettings _validitySettings = validitySettings.Value;
```

`_validitySettings` has no callers after we remove this line, so I think we need to remove that and `IOptions<IdProofingValiditySettings> validitySettings`.  I don't think it's giving any build warnings (which this PR helps deal with a subset of) so it's not a blocker, but it might be a good idea to remove it.

noxmwalsh: Addendum: I'm actually surprised it's not giving a warning for unreadable constructor params at build time, but 🤷🏻

noxmwalsh: Oh, it's technically being read in the field initializer (see **validitySettings.Value**):

```
    private readonly IdProofingValiditySettings _validitySettings = validitySettings.Value;
```

I guess it wouldn't follow through to make sure that _validitySettings has a caller.

noxmwalsh (approved):
Approved with comments

src/SEBT.Portal.Infrastructure/Repositories/DatabaseUserRepository.cs
jamesmblair: That's a good flag. There are some style enforcements like this we could trigger additional warnings/errors on using various analyzer nuget packages. I'll go ahead and take care of this though.

PR #410, opened 2026-06-17
DC-367: Build Redis connection from structured settings, with legacy fallback

jamesmblair:
## Jira ticket
[DC-367](https://codeforamerica.atlassian.net/browse/DC-367)

## Description
Replaces the raw `ConnectionStrings:Redis` connection string with a typed `RedisSettings` class (`Host`, `Port`, `Password`, `Ssl`, `SslHost`) that maps directly to AWS Elasticache configuration. Structured settings take precedence; `ConnectionStrings:Redis` still works as a fallback so existing deployments aren't broken.

Also wires the existing Redis service in `compose.yaml` to the API container for local docker testing.

## Related PRs
None

## Changes
- **`RedisSettings`** — new typed settings class in `Core/AppSettings/`
- **`Dependencies.cs`** — `ResolveRedisConfigurationOptions()` private helper encapsulates precedence logic (structured → legacy connection string → null); both `AddCaching` and `AddDistributedLocking` consume it
- **`appsettings.co.example.json`** — updated to show the new structured format with TLS comments for Elasticache
- **`compose.yaml`** — API container now connects to the Redis service with a health-check dependency
- **`PortalWebApplicationFactory`** — adds `Redis__Host` to the env-var cleanup list alongside the legacy `ConnectionStrings__Redis`
- **`DependenciesTests`** — 6 new unit tests covering all three resolution branches

> ⚠️ **IaC note for reviewers**: The new `Redis:Host`, `Redis:Port`, `Redis:Ssl`, and `Redis:SslHost` config keys will need to be plumbed through Tofu/Secrets Manager for CO production. The legacy `ConnectionStrings:Redis` fallback means nothing breaks until that wiring lands, but it should be tracked.

## Checklist
- [x] Tests added for new behavior (6 unit tests covering all resolution branches)
- [x] Backwards compatible — `ConnectionStrings:Redis` still works
- [x] No hardcoded secrets
- [x] `appsettings.co.example.json` updated with new key format and TLS guidance
- [ ] IaC wiring for `Redis:*` keys in CO production (tracked separately)

[DC-367]: https://codeforamerica.atlassian.net/browse/DC-367?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


.claude/settings.json
```diff
@@ -1,4 +1,7 @@
 {
```
jamesmblair, line 1: Tweaks made based on things encountered during feature development

src/SEBT.Portal.Infrastructure/Configuration/AppConfigAgentReloadService.cs
```diff
@@ -48,14 +48,18 @@ protected override async Task ExecuteAsync(CancellationToken stoppingToken)
             return;
         }
 
+        // Create the timer before logging "started": the test waits for that log message and
```
jamesmblair, line 51: Fix for test flakiness encountered during testing this feature

spokenbird: Thanks for catching this!

src/SEBT.Portal.Infrastructure/Dependencies.cs
```diff
@@ -260,6 +262,56 @@ public static IServiceCollection AddDistributedLocking(
...
+    /// Returns null when neither is configured.
+    /// </summary>
+    internal static ConfigurationOptions? ResolveRedisConfigurationOptions(
+        IConfiguration? configuration,
+        IHostEnvironment environment)
+    {
+        var settings = configuration?.GetSection(RedisSettings.SectionName).Get<RedisSettings>();
+        if (settings?.IsConfigured == true)
+        {
+            if (settings.AcceptSelfSignedCertificates && environment.IsProduction())
```
spokenbird: Not sure CO how names their lower environments, but I imagine they'd want to use the AWS certs there too. Should this actually be `!environment.IsDevelopment()` instead?

test/SEBT.Portal.Tests/Unit/Infrastructure/DependenciesTests.cs
spokenbird, line 1: Test coverage is good, but I think it'd be nice to test the `AcceptSelfSignedCertificates` functionality and TLS mapping as well.

scripts/dev/gen-redis-certs.sh
spokenbird, line 1: Some thoughts here.

We could wire up a guard script, something along the lines of:
```
# scripts/dev/check-redis-certs.sh
#!/bin/bash
if [ ! -f certs/redis-server.crt ]; then
  echo "❌ Redis TLS certs are missing." >&2
  echo "   Run: ./scripts/dev/gen-redis-certs.sh" >&2
  exit 1
fi
```

and wire this up with a predev hook. This is an easy guard and let's people know what's wrong when they try to start things up after this merges.

A more robust option would be some kind of init step that automatically runs the generate cert script for local dev. This removes any need for the developer to run the script themselves before their first startup and removes the head scratch of why things aren't working. That said, this is mostly resolved by the easier suggestion above, unless someone explicitly runs `docker compose up -d` themselves. This could be good follow up work? Thoughts?

spokenbird (changes requested):

scripts/dev/gen-redis-certs.sh
jamesmblair: Yeah a guard script would be easy enough - I can do that. For the more robust and automatic "init step", did you have something specific in mind? Like when would that run and what would trigger it?

package.json
```diff
@@ -36,7 +36,10 @@
     "ci:validate": "./act-test.sh dry-run",
     "check:cooldown-exclusions": "tsx scripts/security/check-cooldown-exclusions.ts",
     "test:scripts": "tsx --test scripts/**/*.test.ts",
-    "prepare": "husky"
+    "prepare": "husky",
+    "claude:api:build": "dotnet build --nodeReuse:false -maxcpucount:1",
```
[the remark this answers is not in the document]
jamesmblair: @AikidoSec ignore:  I think this is a false positive — these scripts are convenience aliases for dotnet CLI commands and don't execute any pnpm dependency management tasks.

scripts/dev/format-staged.sh
```diff
@@ -0,0 +1,27 @@
+#!/bin/sh
```
jamesmblair, line 1: New helper script for running `dotnet format` only on staged files

.husky/pre-commit
```diff
@@ -116,24 +116,27 @@ if [ "$BACKEND_CHANGES" = "true" ]; then
...
-    }
-
-    dotnet format SEBT.Portal.sln analyzers --verify-no-changes --verbosity quiet \
-      --exclude ../sebt-self-service-portal-state-connector/ \
-      --exclude ../sebt-self-service-portal-dc-connector/ \
-      --exclude ../sebt-self-service-portal-co-connector/ || {
-      log_error ".NET analyzer checks failed - fix the reported issues"
-      exit 1
-    }
+    # Verify formatting only for staged .cs files. --include scopes dotnet format to
```
jamesmblair, line 119: Updates to precommit to run `dotnet format` only on staged files

spokenbird (approved):

PR #433, opened 2026-06-26
DC-528 - Add ElastiCache Valkey for shared Redis cache

spokenbird:
#### 🔗 Jira ticket
DC-528

#### ✍️ Description
 Provisions a shared Redis cache (ElastiCache for Valkey 7.2) in both the DC and CO
 dev environments via a new `sebt_redis` OpenTofu module, and wires the API container
 to it.
  
**Note**: This PR explicitly does not scale CO back to 2 container instances `desired_containers = 2` which will be needed to fully exercise the`missing_session` / OIDC login fix that implementing Redis should solve for CO. That can be done as a single line change follow up PR once this PR lands. 

#### 🔗 Links to related PRs
- #410 (DC-367) — defines the `RedisSettings` schema this module's env vars populate. **#410 (DC-367) MUST MERGE BEFORE THIS PR (#433)

**What's included:**
- New `tofu/modules/sebt_redis` module (mirrors `sebt_database`):
- `aws_elasticache_replication_group` — Valkey 7.2, cluster-mode disabled,
  1 primary + 1 replica (`cache.t4g.micro`), Multi-AZ with automatic failover.
- Encryption in transit (TLS) and at rest (customer-managed KMS key).
- AUTH token generated via `random_password`, stored in Secrets Manager
  (encrypted with the module's KMS key), `auth_token_update_strategy = "ROTATE"`.
- Security group allowing ingress on 6379 only from the API service's security group.
- `sebt_application` wiring:
- Injects `Redis__Host`, `Redis__Port`, `Redis__Ssl`, `Redis__SslHost` as env vars
  and `Redis__Password` as a secret (`<auth_token_secret_arn>:auth_token`).
- These map to the `RedisSettings` schema introduced in #410.
- Applied identically to DC and CO. Validated with `tofu plan` against both envs
  (Redis-only diff when pinned to the deployed image tag).

#### ✅ Completion tasks

- [X] Added relevant tests
- [X] Meets acceptance criteria in ticket
- [X] Configuration changes:
  - [X] If new environment variables are added, update in Tofu
  - [ ] If new environment secrets are added, update in Tofu and set in AWS Secret Manager (TO DO AFTER MERGING)


tofu/modules/sebt_redis/local.tf
```diff
@@ -0,0 +1,27 @@
+locals {
+  prefix        = "${var.project}-${var.environment}"
+  project_short = var.project_short != "" ? var.project_short : var.project
+  short_prefix  = "${local.project_short}-${var.environment}"
+  port          = 6379
```
jamesmblair, line 5: For TLS, this should be 6380

tofu/modules/sebt_redis/variables.tf
```diff
@@ -0,0 +1,107 @@
...
+variable "project_short" {
+  type        = string
+  description = "Abbreviated project name for resource naming."
+  default     = ""
+}
+
+variable "secret_recovery_period" {
+  type        = number
+  description = "Number of days before the AUTH token secret is permanently deleted after destruction. 0 forces immediate deletion (dev-friendly)."
+  default     = 0
```
jamesmblair: Is there risk to immediate deletion? Seems like low harm to set this to small but non-zero value. 3 days? 7 days?

tofu/modules/sebt_redis/main.tf
```diff
@@ -0,0 +1,86 @@
...
+  transit_encryption_enabled = true
+  auth_token                 = random_password.auth_token.result
+  auth_token_update_strategy = "ROTATE"
+
+  snapshot_retention_limit = var.snapshot_retention_limit
+  apply_immediately        = var.apply_immediately
+
+  tags = {
+    Name = "${local.prefix}-redis"
+  }
```
jamesmblair, line 85: Thoughts on adding the below item here to validate across these two variables?

```terraform
lifecycle {
  precondition {
    condition     = !var.automatic_failover_enabled || var.num_cache_clusters >= 2
    error_message = "automatic_failover_enabled requires num_cache_clusters >= 2."
  }
}
```

jamesmblair (changes requested):

tofu/modules/sebt_redis/local.tf
spokenbird: StackExchange.Redis defaults to 6380 when SSL is enabled and no port is specified. Your local compose in #410 splits plaintext on 6379 and TLS on 6380 but as I understand it ElastiCache doesn't split ports that way. [ElastiCache defaults to 6379 ](https://docs.aws.amazon.com/AmazonElastiCache/latest/dg/accessing-elasticache.html#:~:text=How%20to%20access%20ElastiCache%20resources%20from%20outside%20AWS)(with or without TLS enabled) and since `ResolveRedisConfigurationOptions` from your #410 passes the port explicitly StackExchange.Redis default should never kick in. [ElastiCache serves TLS on 6379](https://docs.aws.amazon.com/AmazonElastiCache/latest/dg/connect-tls.html#:~:text=To%20connect%20to%20a%20cluster%20with%20encryption%20and%20authentication%20enabled%2C%20enter%20this%20command%3A) and since we pass that explicitly that's what it should connect to.

tofu/modules/sebt_redis/variables.tf
spokenbird: Good catch, thanks!

tofu/modules/sebt_redis/versions.tf
```diff
@@ -0,0 +1,15 @@
+terraform {
+  required_version = ">= 1.6.0"
+
+  required_providers {
+    aws = {
```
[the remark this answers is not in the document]
spokenbird: Will fix in a separate follow up PR

tofu/modules/sebt_redis/versions.tf
```diff
@@ -0,0 +1,15 @@
...
+      source = "hashicorp/aws"
+      # Valkey engine support for aws_elasticache_replication_group landed in 5.73.0.
+      version = ">= 5.73"
+    }
+    random = {
```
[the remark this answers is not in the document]
spokenbird: Will fix in a separate follow up PR

jamesmblair (approved):

PR #451, opened 2026-07-07
DC-558 - Enrollment Checker Outage page + configuration

necampanini:
#### 🔗 Jira ticket
https://codeforamerica.atlassian.net/browse/DC-558

#### ✍️ Description
When a scheduled maintenance window or the manual outage flag takes the backend down, the enrollment checker now redirects every route to the same full-page outage screen the portal shows, and returns users to the form once the outage ends. Outage schedule windows gain an optional `Target` (`Portal`, `EnrollmentChecker`, or `Both`; omitted means `Both`), and a new `checker_outage_page_enabled` flag covers unscheduled emergencies when no windows are scheduled. The checker picks up outage state through its existing features poll, so outages can be turned on and off at runtime via AWS AppConfig without redeploying the statically-hosted app.

#### 🔗 Links to related PRs
None (all changes are in this repo).

#### ✅ Completion tasks

- [x] Added relevant tests
- [x] Configuration changes:
  - [ ] If new environment variables are added, update in Tofu — n/a, no new env vars
  - [ ] If new environment secrets are added, update in Tofu and set in AWS Secret Manager — n/a, no new secrets
  - [x] If you're adding an appsetting, add it to the requisite example file, and update it in the AWS AppConfig — added to `appsettings.json` + both state example files; AWS AppConfig still needs the key added per state (see below)
  - [x] If appsetttings are changed, update in AWS AppConfig

#### ⚙️ Config update needed
- New appsetting `FeatureManagement:checker_outage_page_enabled` (bool, default `false`) is in `appsettings.json` and both state example files. Add the same key to the AWS AppConfig AppSettings profile for DC and CO before merge so runtime toggling works immediately.
- `OutageSchedule:Windows[].Target` (optional string: `Portal` / `EnrollmentChecker` / `Both`; omitted = `Both`) is a new field on the existing `OutageSchedule` block already stored in AppConfig. 
- No new top-level key to add, but we should let state partners know `Target` is available.

---

### Priority Review Table

| Priority | File | Unit | Sec | Cpx | Nov | Notes |
|:---:|---|---|:---:|:---:|:---:|---|
| 🟡 | `packages/design-system/content/scripts/generate-locales.js` | `filterStateDataForApp()` | Low | Med | Med | per-app namespace filtering |
| 🟡 | `packages/design-system/src/components/OutagePage/OutagePageContent.tsx` | `OutagePageContent` | Low | Low | Med | shared via new `resources` prop |
| 🟡 | `src/SEBT.EnrollmentChecker.Web/src/app/layout.tsx` | `RootLayout` | Low | Low | Med | chrome moves to client shell |
| 🟡 | `src/SEBT.EnrollmentChecker.Web/src/components/OutageGuard.tsx` | `OutageGuard` | Med | Med | Med | redirect + cached-state gating |
| 🟡 | `src/SEBT.EnrollmentChecker.Web/src/features/outage/outageFlagCache.ts` | cache helpers | Med | Low | Low | sessionStorage, boolean only |
| 🟡 | `src/SEBT.EnrollmentChecker.Web/src/features/outage/useOutageState.ts` | `useOutageState()` | Low | Med | Med | stale-poll policy inverts banner's |
| 🟡 | `src/SEBT.Portal.Api/Controllers/EnrollmentCheck/EnrollmentCheckController.cs` | `GetFeatures()` | Med | Low | Low | public endpoint, new field |
| 🟡 | `src/SEBT.Portal.Infrastructure/Services/FeatureFlagQueryService.cs` | `ApplySurfaceOutageFlag()` | Med | Med | Low | portal flag now per-surface |
| 🟡 | `src/SEBT.Portal.Infrastructure/Services/OutagePageStateResolver.cs` | `IsOutagePageActiveAsync()` | Med | Med | Med | schedule-wins-per-surface rule |
| 🟡 | `src/SEBT.Portal.Infrastructure/Services/OutageScheduleEvaluator.cs` | `HasScheduledWindows()` / `TryParseTarget()` | Med | Med | Low | bad-date windows still count |
| 🟢 | `packages/design-system/src/components/OutagePage/getFooterLinkLabel.ts` | move from portal + tests | Low | Low | Low | |
| 🟢 | `packages/design-system/src/components/OutagePage/getOutageMessages.ts` | move from portal + tests | Low | Low | Low | |
| 🟢 | `packages/design-system/src/index.ts` | barrel exports | Low | Low | Low | |
| 🟢 | `src/SEBT.EnrollmentChecker.Web/package.json` | dependency entry | Low | Low | Low | |
| 🟢 | `src/SEBT.EnrollmentChecker.Web/src/app/outage/page.tsx` | `OutagePage` | Low | Low | Low | |
| 🟢 | `src/SEBT.EnrollmentChecker.Web/src/components/CheckerShell.tsx` | `CheckerShell` | Low | Low | Low | |
| 🟢 | `src/SEBT.EnrollmentChecker.Web/src/features/maintenance/api/fetchCheckerFeatures.ts` | schema: optional `outagePage` | Low | Low | Low | |
| 🟢 | `src/SEBT.EnrollmentChecker.Web/src/lib/analytics-routes.ts` | `/outage` route entry | Low | Low | Low | |
| 🟢 | `src/SEBT.EnrollmentChecker.Web/src/mocks/handlers.ts` | mock features payload | Low | Low | Low | |
| 🟢 | `src/SEBT.EnrollmentChecker.Web/src/**/*.test.{ts,tsx}` | frontend tests (5 files) | Low | Low | Low | |
| 🟢 | `src/SEBT.Portal.Api/Models/.../EnrollmentCheckerFeaturesResponse.cs` | `OutagePageFeature` DTO | Low | Low | Low | |
| 🟢 | `src/SEBT.Portal.Api/appsettings.json` + `appsettings.{co,dc}.example.json` | config blocks + docs | Med | Low | Low | |
| 🟢 | `src/SEBT.Portal.Core/AppSettings/FeatureFlags.cs` | `CheckerOutagePageEnabled` constant | Low | Low | Low | |
| 🟢 | `src/SEBT.Portal.Core/AppSettings/OutageScheduleSettings.cs` | `Target` property, `OutageTarget` enum | Low | Low | Low | |
| 🟢 | `src/SEBT.Portal.Infrastructure/Dependencies.cs` | DI registrations | Low | Low | Low | |
| 🟢 | `src/SEBT.Portal.Infrastructure/Services/IOutagePageStateResolver.cs`, `IOutageScheduleEvaluator.cs` | interfaces | Low | Low | Low | |
| 🟢 | `src/SEBT.Portal.Web/src/app/(public)/outage/page.tsx` | consume shared component | Low | Low | Low | |
| 🟢 | `src/SEBT.Portal.Web/src/features/outage/*` | deletions (moved to design system) | Low | Low | Low | |
| 🟢 | `test/SEBT.Portal.Tests/**` | backend tests (4 files) | Low | Low | Low | |


src/SEBT.Portal.Infrastructure/Services/OutageScheduleEvaluator.cs
```diff
@@ -74,6 +91,41 @@ public bool IsOutageActive()
         return false;
     }
 
+    public bool HasScheduledWindows(OutageTarget surface)
```
spokenbird, line 99: Since IsOutageActive checks the date being malformed but HasScheduledWindows here does not there is an issue where a malformed date would log a warning, but it would lock the out the outage page all together both from the scheduler and from the explicit flag. 

I think we could extract out the validation check to a helper method and check in both places which would allow the explicit flag to trump a malformed date. Otherwise I worry someone setting a malformed date may not notice the log and permanently lock out the functionality until they notice the malformed date. 

EDITED from discussion in standup:

We should up the LogWarn to LogError and we could use Options validation to check for malformed dates or targets and fail fast: https://learn.microsoft.com/en-us/aspnet/core/fundamentals/configuration/options?view=aspnetcore-10.0

src/SEBT.Portal.Infrastructure/Services/OutagePageStateResolver.cs
```diff
@@ -0,0 +1,78 @@
...
+        {
+            OutageTarget.Portal => FeatureFlags.OutagePageEnabled,
+            OutageTarget.EnrollmentChecker => FeatureFlags.CheckerOutagePageEnabled,
+            _ => throw new ArgumentOutOfRangeException(
+                nameof(surface),
+                surface,
+                "Pass a single surface (Portal or EnrollmentChecker); Both is a window-level value.")
+        };
+
+        if (_scheduleEvaluator.HasScheduledWindows(surface))
```
spokenbird, line 39: This functionality of schedule trumps feature flag is duplicated in FeatureFlagQueryService ApplySurfaceOutageFlag we should consider DRYing it up so one or the other is the authority / future drift doesn't happen?

spokenbird (changes requested):
Over all looks good but a few things I call out. 

Additionally I noticed that OutagePageContent, getOutageMessages and getFooterLinkLabel were moved into packages/design-system and I wonder if it makes sense to do something similar with the duplicated logic across outageFlagCache, OutageGuard could do similar? Not sure if that's a good idea as I have less frontend knowledge here, but seems like a lot of duplication? Curious your thoughts.

src/SEBT.Portal.Infrastructure/Services/OutageScheduleEvaluator.cs
necampanini: Addressed in 00841a10

src/SEBT.Portal.Infrastructure/Services/OutagePageStateResolver.cs
necampanini: Addressed in 00841a10

spokenbird (approved):
Changes look good, thanks for being so thorough!

PR #458, opened 2026-07-10
[DC-505] Capture core Web Vitals on page load via SEBT data layer

adbergen:
#### 🔗 Jira ticket
https://codeforamerica.atlassian.net/browse/DC-505

#### ✍️ Description
Captures the five core Web Vitals (TTFB, FCP, LCP, CLS, INP) via the [web-vitals](https://github.com/GoogleChrome/web-vitals) library and reports each through the SEBT data layer as a `web_vitals` event.

**Event shape** (flat snake_case, one event per metric as it finalizes, at most five per hard page load):
- `metric_name` — `ttfb` | `fcp` | `lcp` | `cls` | `inp`
- `metric_value` — integer milliseconds for timing metrics; unitless 4-decimal for CLS
- `metric_rating` — `good` | `needs-improvement` | `poor`
- `page_instance_id` / `initial_path` — pin late-finalizing metrics (CLS/INP report at page-hide) to the hard page load they measured, since the auto-snapshotted page context may reflect a later soft-navigated page

**Design notes:**
- Metrics finalize at different times (CLS/INP only at page-hide), so they can't ride on the `page_load` event itself — the per-metric event keyed by `page_instance_id` is the ticket's alternative approach.
- CLS/INP are cumulative and can grow after first finalizing; the recorded value is the metric as of the first page-hide — an accepted tradeoff for strict once-per-load semantics (documented in the `WEB_VITALS` doc comment).
- The adapter lives in `packages/analytics` and is auto-mounted by `DataLayerProvider`, so **both the portal and the enrollment checker** report vitals with no per-app wiring. Existing vendor bridges (Amplitude/Mixpanel/SiteImprove) forward the events unchanged.
- No CSP change needed: metrics are collected locally via PerformanceObserver, no new external domain.
- Unsupported browsers (e.g. Safari for LCP/CLS/INP) simply never fire those callbacks — no errors.

**Follow-up for the analytics team:** the new `web_vitals` event and its fields need adding to the SEBT Data Layer Dictionary.

#### 🔗 Links to related PRs
None.

#### ✅ Completion tasks

- [x] Added relevant tests (adapter unit tests incl. rounding/dedupe/StrictMode; provider mount tests; both app suites green)
- [x] Meets acceptance criteria in ticket (fields + units per schema above; manual Amplitude verification on a lower env pending deploy — when doing it, close the tab for real so CLS/INP delivery through the vendor SDK is confirmed)
- [x] Configuration changes: none
  - [x] No appsettings added
  - [x] No appsettings changed

---

#### 🧪 QA Summary

**What changed:** No visible UI changes — the portal and enrollment checker now report five performance metrics as `web_vitals` analytics events. Verification is via browser console and Amplitude.

**Happy path**
1. Open any page (signed-out landing page works). In the dev console run `window.digitalData.event.filter(e => e.eventName === 'web_vitals')` — `ttfb` and `fcp` appear shortly after load.
2. Click the page, switch tabs and back — `lcp` appears; after hiding the tab, `cls` and `inp` too (max 5 entries).
3. Check payloads: `metric_value` is whole-number ms except `cls` (small decimal); `metric_rating` ∈ good/needs-improvement/poor; all entries share one `page_instance_id` + `initial_path`.

**Edge cases**
1. Client-side navigation produces no new vitals batch; `initial_path` stays the originally loaded page. A refresh starts a new batch with a new `page_instance_id`.
2. Hiding/showing the tab repeatedly never duplicates a metric.
3. Safari: only `ttfb`/`fcp` may appear — no console errors.
4. Same behavior on the enrollment checker.

**Regression**
1. `page_load` still fires per navigation; `cta_click` still tracked.
2. On an Amplitude-enabled env: load, interact, close the tab entirely → `web_vitals` events (incl. `cls`/`inp`) arrive and are queryable by `metric_name`.

**Credentials:** console checks need no sign-in; for the signed-in check on DC local/dev use `sebt.dc+co-loaded@codeforamerica.org` (OTP via Mailpit at http://localhost:8025 locally).


packages/analytics/src/web-vitals.ts
```diff
@@ -0,0 +1,65 @@
...
+
+  function report(metric: Metric): void {
+    if (disposed || reported.has(metric.name)) return
+    reported.add(metric.name)
+
+    dl.trackEvent(WEB_VITALS, {
+      metric_name: metric.name.toLowerCase(),
+      metric_value: normalizeValue(metric),
+      metric_rating: metric.rating,
+      page_instance_id: pageInstanceId,
```
mtodd-cfa: I think instead of trying to stitch this together with the path and generated instance id, we can use the `DataLayer#get()` method to populate some of the same values we are also sending with i.e. pageload. I'm thinking:

- page.name
- page.language
- page.environment
- page.application
- page.flow
- page.step

mtodd-cfa (changes requested):

mtodd-cfa (approved):

PR #462, opened 2026-07-14
[DC-583] Invalidate portal JWT on logout

adbergen:
#### 🔗 Jira ticket
https://codeforamerica.atlassian.net/browse/DC-583

#### ✍️ Description
Logout now revokes the portal JWT server-side. Previously, a token copied before logout (from a Bearer header, log, or compromised client) stayed usable against authenticated endpoints until `exp`.

**Mechanism:**
- New `ITokenDenylist`/`TokenDenylist` (Api layer, beside `PreAuthSessionStore`) stores `auth:denylist:{jti}` in `IDistributedCache` (Redis in deployed envs, memory fallback in dev) with TTL = token remaining lifetime + the bearer clock skew — entries self-expire, no unbounded growth.
- `GET /api/auth/logout` best-effort denylists the surrendered cookie token's jti (parsed without validation — the token is being discarded). Missing cookie / malformed token / missing jti → logout proceeds exactly as before. Cookie clearing and the OIDC end-session → `/login` redirect chain are unchanged.
- The JwtBearer `OnTokenValidated` event rejects denylisted jtis on every authenticated request (after the existing `SessionLifetimePolicy` check), for both cookie and `Authorization: Bearer` presentations.

**Security posture (deliberate choices):**
- **Fail open**: cache errors are logged and treated as not-denylisted, and a failed denylist write never fails logout — a Redis blip degrades revocation, never availability.
- **Hardened against forged input**: logout is anonymous and parses unvalidated tokens, so the jti must parse as a Guid (the only shape the portal mints) and the TTL is clamped to `JwtSettings.ExpirationMinutes` + clock skew — a forged far-future `exp` cannot create long-lived cache entries.
- The bearer `ClockSkew` and the denylist padding now share one constant, so they cannot silently drift apart.

**Follow-ups for the team (surfaced by security review, out of scope here):**
1. `/api/auth/refresh` rotates to a new jti without denylisting the old one — a stolen token can be "refreshed past" a later logout within the 60-minute absolute session cap. Suggest a follow-up ticket to denylist the surrendered jti on refresh (trivial now that `ITokenDenylist` exists).
2. Verify DC prod topology: without a Redis config the denylist runs on per-process memory cache (entries lost on app-pool recycle, not shared across nodes). CO (Redis-backed) is unaffected.

#### 🔗 Links to related PRs
None.

#### ✅ Completion tasks

- [x] Added relevant tests (15 unit + 8 integration through the real pipeline: revocation on both presentations, TTL clamp, forged-jti rejection, fail-open both directions, OIDC + non-OIDC redirects, user-B isolation)
- [x] Configuration changes: none
  - [x] No appsettings added or changed (reuses existing `JwtSettings` and Redis config)

---

#### 🧪 QA Summary

**What changed:** No visible UI change — signing out now server-side revokes the session token, so a copied token no longer works after logout.

**Happy path**
1. Sign in (DC), copy the `sebt_portal_session` cookie value from dev tools.
2. Sign out — login page, cookie gone.
3. Replay the copied value (dev tools re-add, or `curl -H "Cookie: sebt_portal_session=<value>" .../api/auth/status`) → 401 / treated as signed out.

**Edge cases**
1. User A's logout leaves user B's session (other browser) working.
2. Logout with no session → login page, no error.
3. CO env: logout still routes through the MyColorado end-session page back to login.

**Regression:** sign-in end-to-end; session refresh keeps an active user signed in.

**Credentials:** DC local/dev `sebt.dc+co-loaded@codeforamerica.org` (+ any second seeded persona for the two-user test), OTPs via Mailpit at http://localhost:8025 locally.


src/SEBT.Portal.Api/Services/TokenDenylist.cs
```diff
@@ -0,0 +1,96 @@
...
+        {
+            return await _cache.GetAsync(CacheKeyPrefix + jti, cancellationToken) is not null;
+        }
+        catch (OperationCanceledException) when (cancellationToken.IsCancellationRequested)
+        {
+            throw;
+        }
+        catch (Exception ex)
+        {
+            _logger.LogError(ex, "Token denylist lookup failed; failing open");
```
necampanini: with`OnTokenValidated` running on every on every authenticated request, I think something like a Redis outage could flood our logging. maybe throttling or sampling could help limit it

apps/portal/test/SEBT.Portal.Tests/Unit/Controllers/AuthControllerTests.cs
necampanini, line 1: good coverage, but we could add a test asserting `DenyAsync` is called with the parsed `jti` and `ValidTo`?

necampanini (changes requested):
looking good, just minor changes requested

necampanini (approved):
requested changes addressed and its looking good, approved!

PR #466, opened 2026-07-17
[DC-602] Disable household data refetch on window focus

adbergen:
#### 🔗 Jira ticket
https://codeforamerica.atlassian.net/browse/DC-602

#### ✍️ Description
Removes the `refetchOnWindowFocus: true` opt-in from `useHouseholdData` — previously, every tab switch back to the portal fired `/household/data` (twice in Colorado's deferred-card mode), fanning out to state agency APIs and risking rate limiting.

**Investigation findings:**
- The flag dates to the dashboard's original commit (PR #41, DC-85) as part of the initial "always fresh" posture alongside `staleTime: 0`; no later work depends on it.
- Freshness is preserved: `staleTime: 0` + default refetch-on-mount means every page navigation still fetches fresh data, which covers the original custody-staleness requirement.
- The app-wide `QueryProvider` default is already `refetchOnWindowFocus: false`; this hook was the only opt-in.
- Session expiry does not rely on it (`TokenRefresher`/`useSessionRefresh` timers own that). One edge-case change: a session that expires while the user is in another tab now redirects to `/login` on their next interaction rather than instantly on refocus.

Both queries now set `refetchOnWindowFocus: false` explicitly (with a comment on the state-API cost) rather than inheriting the provider default, so the behavior is pinned at the hook even if the provider default changes. A new red-first test flips `focusManager` against a bare `QueryClient` (which carries React Query's library default of `true`) and asserts no refetch occurs.

#### 🔗 Links to related PRs
N/A

#### ✅ Completion tasks

- [x] Added relevant tests
- [ ] Configuration changes:

### DIFF
--- a/apps/portal/src/SEBT.Portal.Web/src/features/household/api/useHouseholdData.test.tsx
+++ b/apps/portal/src/SEBT.Portal.Web/src/features/household/api/useHouseholdData.test.tsx
@@ -7,7 +7,7 @@
  * - Custom retry logic (no retry on 4xx, retry on 5xx)
  * - Exponential backoff retry delay
  */
-import { QueryClient, QueryClientProvider } from '@tanstack/react-query'
+import { focusManager, QueryClient, QueryClientProvider } from '@tanstack/react-query'
 import { renderHook, waitFor } from '@testing-library/react'
 import { http, HttpResponse } from 'msw'
 import { afterEach, beforeEach, describe, expect, it, vi } from 'vitest'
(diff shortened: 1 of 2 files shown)


apps/portal/src/SEBT.Portal.Web/src/features/household/api/useHouseholdData.test.tsx
```diff
@@ -356,6 +356,40 @@ describe('useHouseholdData', () => {
...
+      await waitFor(() => {
+        expect(result.current.isSuccess).toBe(true)
+      })
+      expect(fetchCount).toBe(1)
+
+      focusManager.setFocused(false)
+      focusManager.setFocused(true)
+
+      // Give any wrongly-triggered refetch time to hit the mock server
+      await new Promise((resolve) => setTimeout(resolve, 50))
```
noxmwalsh, line 389: I don't think there's any other promise that's returned that we can hook onto for this, so a `setTimeout` make sense here.  I know there's fake timer libs as part of Vitest, but I think those are for advancing timers within the implementing code

apps/portal/src/SEBT.Portal.Web/src/features/household/api/useHouseholdData.test.tsx
```diff
@@ -356,6 +356,40 @@ describe('useHouseholdData', () => {
...
+        )
+      })
+
+      await waitFor(() => {
+        expect(result.current.isSuccess).toBe(true)
+      })
+      expect(fetchCount).toBe(1)
+
+      focusManager.setFocused(false)
+      focusManager.setFocused(true)
```
noxmwalsh, line 386: Handy!

noxmwalsh (approved):
I think this looks good to me

PR #467, opened 2026-07-17
[DC-471] Send guardian email and Socure identifier on DC co-loaded household fetch

adbergen:
#### 🔗 Jira ticket
https://codeforamerica.atlassian.net/browse/DC-471

#### ✍️ Description
DC's data team (Carl) needs two more pieces of data in the `@guardianIdentifiers` JSON to correlate warehouse calls: the guardian's login email and the Socure identifier from identity verification. This PR is the portal/contract side:

- **Contract** (`ISummerEbtCaseService.GetHouseholdByBenefitIdentifierAndDobAsync`): new optional `socureReferenceId` parameter; `StateConnectorInterfacesVersion` bumped 0.0.3 → 0.0.4. CO connector and API default implementations updated (no-ops — only DC consumes it).
- **New lookup**: `IDocVerificationChallengeRepository.GetLatestSocureReferenceIdByUserIdAsync` — projection returning the newest challenge's `SocureReferenceId` (skips the full domain mapping, which would decrypt PII fields this caller never needs).
- **Handler**: the co-loaded IC+DOB fallback in `GetHouseholdDataQueryHandler` resolves the reference id and threads it through `HouseholdRepository` to the plugin. Passed as null when the user never reached a Socure session — the DC connector then omits the key entirely.
- The guardian email already flowed to the DC connector on this path; the connector-side PR moves it into the JSON as well. Non-co-loaded lookups are unchanged.

Resulting JSON DC receives on the co-loaded fetch:
```json
{ "PortalUUID": "…", "IC": "…", "DOB": "1984-03-05", "Email": "guardian@example.com", "SocureUUID": "a7b8c9d0-…" }
```

Tests: handler theory covering present-and-null Socure id flows; Testcontainers tests for the new repository lookup (newest-wins, none-present, cross-user isolation); existing suites updated for the new signatures.

#### 🔗 Links to related PRs
DC connector (JSON builder): https://github.com/codeforamerica/sebt-self-service-portal-dc-connector/pull/48

#### ✅ Completion tasks

- [x] Added relevant tests
- [ ] Configuration changes:


noxmwalsh:
Confirmed via .NET logs that we're sending this, so this is effectively Acceptance Test approved.  I'm still following up on code review, so stay tuned

apps/portal/test/SEBT.Portal.Tests/Unit/Repositories/HouseholdRepositoryTests.cs
```diff
@@ -151,6 +151,7 @@ public async Task GetHouseholdByBenefitIdentifierAndGuardianDobAsync_DelegatesTo
                 Arg.Any<PluginPiiVisibility>(),
                 PluginIdentityAssuranceLevel.IAL1plus,
                 userId,
+                Arg.Any<string?>(),
```
noxmwalsh, line 156: Can we update the tests in this file to pass an explicit `socureReferenceId` (and ideally a null case via [Theory]) into `GetHouseholdByBenefitIdentifierAndGuardianDobAsync`and assert that exact value on the plugin call?

This is just to make sure it's passing through as expected.  `Any` will, I think, fall back to `null` (which is valid too in this case, but to test the use case of the work)

noxmwalsh (changes requested):
@adbergen Additional unit test requested, but the functionality works as expected.  Is it possible if we can cover the case mentioned in comment? 

Good job otherwise on the PR 👍🏻

noxmwalsh (approved):
Approved

PR #468, opened 2026-07-17
[DC-341] Instrument the web apps with OpenTelemetry

jamesmblair:
#### 🔗 Jira ticket
https://codeforamerica.atlassian.net/browse/DC-341

#### ✍️ Description
First of a two-PR stack for DC-341 — OpenTelemetry on the Next.js web tier.

Adds `@sebt/observability`: a shared, server-only OTEL Node SDK bootstrap. It emits OTLP (traces, metrics, logs) to a collector, configured entirely from standard `OTEL_*` env vars — gRPC by default (matches the backend), `http/protobuf` switchable, console fallback. It's inert when no endpoint is set, so this ships dark. Config resolution is pure and unit-tested (26 tests).

Both Next apps register it via `instrumentation.ts` (Node-runtime guarded). Nothing deployed changes yet — PR B sets the endpoint and turns it on.

Two deliberate gaps:
- The logs pipeline is wired but dark until a pino retrofit gives it an emit source.
- The enrollment checker is instrumented but deploys as a static export, so its server OTEL is a build-time no-op. Kept SSR-ready.

#### 🔗 Links to related PRs
- #469

#### ✅ Completion tasks
- [x] Added relevant tests
- [ ] Configuration changes: none deployed here — the `OTEL_*` vars land in PR B (Tofu + `web.config`)

### DIFF
--- a/packages/observability/src/config.ts
+++ b/packages/observability/src/config.ts
@@ -68,11 +68,26 @@ function parseProtocol(value: string | undefined): OtlpProtocol {
 }
 
 /**
+ * Default mode for a single signal. A signal is exported over OTLP when the
+ * general endpoint is set (which enables all signals) or when its own
+ * signal-specific endpoint is set; otherwise it stays dark ('none'). This keeps
+ * a lone signal-specific endpoint from silently enabling the other two.
+ */
+function signalDefault(
+  hasGeneralEndpoint: boolean,
+  signalEndpoint: string | undefined
+): ExporterMode {
+  return hasGeneralEndpoint || cleanString(signalEndpoint) !== undefined ? 'otlp' : 'none'
+}
+
+/**
  * Resolve OpenTelemetry configuration from the environment.
  *
- * The default per-signal mode is driven by whether any OTLP endpoint is
- * configured: endpoint present -> 'otlp'; absent -> 'none'. "None" is the
- * inert/dark state — e.g. a deployed environment not yet pointed at a collector.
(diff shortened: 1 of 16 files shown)


pnpm-lock.yaml
necampanini, line 1: this was mentioned in stand up but just noting it here: CI was throwing due to something in this file

packages/observability/src/config.ts
```diff
@@ -0,0 +1,135 @@
...
+ * OTEL_{TRACES,METRICS,LOGS}_EXPORTER variables (otlp | console | none), which
+ * is how local dev flips a signal to the console exporter for quick inspection.
+ */
+export function resolveOtelConfig(
+  env: NodeJS.ProcessEnv,
+  options: ResolveOtelConfigOptions
+): OtelConfig {
+  const hasEndpoint =
+    cleanString(env.OTEL_EXPORTER_OTLP_ENDPOINT) !== undefined ||
+    cleanString(env.OTEL_EXPORTER_OTLP_TRACES_ENDPOINT) !== undefined ||
```
necampanini: think there's an edge case here if only the `OTEL_EXPORTER_OTLP_TRACES_ENDPOINT` is set, then down in the config object, `traces`, `metrics` and `logs` all get the `otlp` mode, even if metrics and logs weren't specified

unsure if that's an actual issue though

necampanini (approved):
approving with comments, nothing blocking

PR #495, opened 2026-07-21
DC-250: Provision rotation app-user across all target databases

spokenbird:
#### 🔗 Jira ticket
DC-250

#### ✍️ Description
The DC dev environment's API container has been stuck in a restart loop. Root cause: the credential-rotation Lambda's set_secret/test_secret steps only ever provisioned the app user (`appuser/appuser_clone`) in the `SebtPortal` database. DC also points the DC-connector plugin at a second database on the same RDS instance, `DcSource`, using the same rotating credentials — but nothing ever created a database user for the app login there. Once appuser_clone became the active credential, the API could reach SebtPortal fine but every DcSource connection (including the dc-sql-connectivity health check) failed with SQL error 4060 ("Cannot open database... login failed"), keeping ECS in a permanent restart loop.

This PR:
- Adds `additional_db_names` to the `sebt_database` module and threads it from `sebt_application` as [`var.dc_source_db_name`] — empty for CO, ["DcSource"] for DC, so the Lambda only touches databases that actually exist for a given state.
- Updates `rotate_db_credentials.py` so `set_secret/test_secret` loop over every target database (primary + ADDITIONAL_DB_NAMES) instead of just the primary one.
- Fixes `ecs_service_name` in sebt_application/main.tf, which was aliasing module.api.cluster_name — currently correct only because the CfA fargate module happens to name the ECS cluster and service identically (not because that's a documented contract. Now derived explicitly via a new local.api_ecs_service_name, so a future upstream naming change can't silently break the Lambda's ecs:UpdateService call.
- Adds a distinct, greppable log line (`DB_ROTATION_ECS_RESTART_FAILED`) when that ECS restart call fails, so it can be wired to a Datadog log monitor instead of failing silently.

#### ✅ Completion tasks
- [x] Added relevant tests (new Lambda unit tests for multi-database provisioning, _target_db_names, and per-db connection testing — 18/18 passing)
- [x] Meets acceptance criteria in ticket (restores the "two dedicated SQL Server logins... alternate as the active credential" behavior described in ADR 0018, across every database the app actually needs)
- [x] Configuration changes:
  - [x] If new environment variables are added, update in Tofu (ADDITIONAL_DB_NAMES on the rotation Lambda)
  - [ ] If new environment secrets are added, update in Tofu and set in AWS Secret Manager (n/a)
  - [ ] If you're adding an appsetting, add it to the requisite example file, and update it in the AWS AppConfig (n/a)
  - [ ] If appsettings are changed, update in AWS AppConfig (n/a)

### DIFF
--- a/tofu/modules/sebt_application/locals.tf
+++ b/tofu/modules/sebt_application/locals.tf
@@ -21,6 +21,16 @@ locals {
     DD_API_KEY = data.aws_secretsmanager_secret.datadog_key["this"].arn
   } : {}
 
+  # tofu-modules-aws-fargate-service (module.api) doesn't expose a
+  # service_name output — only cluster_name. Its cluster and service happen
+  # to share the same name today (HENNGE/ecs/aws: join("-", compact([project,
+  # environment, service]))), but that's an implementation detail of the
+  # upstream module, not a contract. Reconstruct it explicitly here so a
+  # future naming-scheme change in that module doesn't silently break the
+  # rotation Lambda's ecs:UpdateService call. If the module ever adds a
+  # service_name output, switch to that instead.
+  api_ecs_service_name = join("-", compact(["${var.project}-${var.state}", var.environment, "api"]))
+
   # Allow the AWS Security Agent penetration test to bypass the WAF in the
   # development environment by matching its unique User-Agent. "allow" is a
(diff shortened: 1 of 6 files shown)


noxmwalsh (approved):
:shipit:

PR #507, opened 2026-07-24
DC-324(Feature): Add RehashEmailHash backfill console app

noxmwalsh:
#### 🔗 Jira ticket
[DC-324](https://codeforamerica.atlassian.net/browse/DC-324)

#### ✍️ Description
Adds a manual ops console (`scripts/RehashEmailHash`) to rewrite `Users.EmailHash` under a chosen `IdentifierHasher:SecretKey` by decrypting stored `Email`

Use this when rotating the identifier/email lookup secret, or healing rows hashed under a different secret than the running API. The tool supports `--dry-run` and `--yes`, skips decrypt failures, and skips collision sets that would violate `IX_Users_EmailHash`. It does **NOT** rehash other IdentifierHasher-backed columns (SSN, card-replacement cooldown hashes, and similar).

#### 🔗 Links to related PRs
N/A

#### ✅ Completion tasks

- [x] Added relevant tests
- [ ] Configuration changes:

[DC-324]: https://codeforamerica.atlassian.net/browse/DC-324?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


apps/portal/src/SEBT.Portal.Infrastructure/Repositories/DatabaseUserRepository.cs
```diff
@@ -373,30 +375,49 @@ public async Task UpdateUserAsync(User user, CancellationToken cancellationToken
...
+        CancellationToken cancellationToken)
+    {
+        if (entity.EmailHash == lookupHash)
+        {
+            return;
+        }
+
+        entity.EmailHash = lookupHash;
+        entity.UpdatedAt = DateTime.UtcNow;
+        await dbContext.SaveChangesAsync(cancellationToken);
```
necampanini: if this reads one of the older rows that still has a plain text email, would it update `EmailHash` but leave `Email` unencrypted? if so, wouldn't the `PiiPlaintextEncryptionBackfill` skip it since it checks for `EmailHash == null`

apps/portal/test/SEBT.Portal.Tests/Unit/Repositories/DatabaseUserRepositoryEmailHashMissTests.cs
```diff
@@ -0,0 +1,97 @@
...
+
+    /// <summary>
+    /// After EncryptAtRest, uniqueness is only on <c>EmailHash</c> while <c>Email</c> is
+    /// non-deterministic AES-GCM ciphertext. If lookup misses the stored MAC, plaintext equality cannot match ciphertext and
+    /// <see cref="DatabaseUserRepository.GetOrCreateUserAsync"/> inserts a second Users
+    /// row for the same address — the old unique index on Email would have blocked that.
+    /// </summary>
+    [Fact]
+    public async Task GetOrCreateUserAsync_WhenEncryptedUserExistsButEmailHashMisses_ShouldReuseExistingUser()
+    {
```
necampanini: good regression test! it catches on `main`. I'd only ask is to add a test that runs through `UpdateUserAsync's` duplicate check

necampanini (changes requested):
looking great, just small changes requested

apps/portal/test/SEBT.Portal.Tests/Unit/Repositories/DatabaseUserRepositoryEmailHashMissTests.cs
noxmwalsh: Sure, I can add too!  Addressed in dff078386ecd826c3b4d4d0a3b420ef9391fd76f

apps/portal/src/SEBT.Portal.Infrastructure/Repositories/DatabaseUserRepository.cs
noxmwalsh: Yeah, that's a good catch.  This was also addressed in dff078386ecd826c3b4d4d0a3b420ef9391fd76f

apps/portal/src/SEBT.Portal.Infrastructure/Repositories/DatabaseUserRepository.cs
```diff
@@ -373,30 +375,49 @@ public async Task UpdateUserAsync(User user, CancellationToken cancellationToken
...
             .ToListAsync(cancellationToken);
 
-        foreach (var orphan in orphans)
+        foreach (var candidate in candidates)
         {
-            if (TryNormalizeStoredEmail(orphan.Email) == normalizedEmail)
+            if (TryNormalizeStoredEmail(candidate.Email) == normalizedEmail)
             {
-                return orphan;
+                return candidate;
```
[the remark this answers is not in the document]
noxmwalsh: The suggestion from Aikido is both incomplete (it would break the app due to syntax errors if we include it) and doesn't make sense to me for what it's trying to accomplish.  It does not reduce or narrow the decrypt count, so I'm not sure if this is worth addressing.

@necampanini Is it fine if we dismiss this comment from Aikido?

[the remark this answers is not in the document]
necampanini: hm yeah, the suggested fix doesn't seem right but I think the underlying find makes sense when/if the `Users` table grows over time. if we don't foresee that many users, the existing scan should be fine as is.

noxmwalsh:
@necampanini Let's hold off on merging this: I'm suspecting we might be over-fixing a problem that's considered an edge case here.

adbergen:
@jhernandez-cfa to look into this to see if duplicate row shows even if encryption is off. This PR may not be needed.

apps/portal/src/SEBT.Portal.Infrastructure/Services/EmailHashRehashService.cs
```diff
@@ -0,0 +1,206 @@
...
+            if (planned.Status == RehashRowStatus.DecryptFailed)
+            {
+                decryptFailures++;
+                continue;
+            }
+
+            plans.Add(planned);
+        }
+
+        var collisionUserIds = FindCollisionUserIds(plans);
```
jhernandez-cfa, line 61: Why are we doing a collision check here on the userId Guid and not the plaintext email? Aren't we trying to avoid a violation of IX_Users_EmailHash?

apps/portal/src/SEBT.Portal.Infrastructure/Repositories/DatabaseUserRepository.cs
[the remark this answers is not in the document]
noxmwalsh: We're no longer doing this and this code was removed, so I'm going to go ahead and resolve this comment

apps/portal/src/SEBT.Portal.Infrastructure/Services/EmailHashRehashService.cs
noxmwalsh: We are grouping by the planned `EmailHash`, not by user Id. `NewHash` is the HMAC of the normalized email (see `PlanRow` around lines 138–149), which is what `IX_Users_EmailHash` indexes.

The actual collision detection is in `FindCollisionUserIds` (specifically lines 103–110, see `GroupBy(p => p.NewHash`) which then keep groups with more than one distinct `UserId`.  The IDs that are being returned and set in `collisionUserIds` is just the list of IDs that need to be deduped.

Let me know if this is an adequate explanation on what's happening here!

PR #509, opened 2026-07-27
[DC-159] Flatten and bring API/state secrets under Doppler management

spokenbird:
#### 🔗 Jira ticket
https://codeforamerica.atlassian.net/browse/DC-159

#### ✍️ Description
Migrates DC and CO's static application secrets (JWT signing key, identifier hasher key, Smarty credentials, Socure credentials, PII encryption keys, CBMS OAuth credentials, MyColorado OIDC credentials) from grouped AWS Secrets Manager entries to individual flat secrets, and syncs them from Doppler via the `tofu-modules-aws-doppler` module — so these can be managed from Doppler going forward instead of the AWS console directly. Also includes:
- Bumped `module.web`'s pinned `tofu-modules-aws-fargate-service` version (1.10.0 → 1.14.0), fixing drift found while testing this change (an ALB TLS policy downgrade and a container-insights setting, both already fixed upstream in versions this pin predated)
- Extracted the previously-hardcoded `project_short` value into a variable
- Added Dependabot tracking for CfA's tofu modules going forward
- Wired `DOPPLER_TOKEN` through to the plan/deploy CI jobs, required for the new Doppler-managed resources

Database and SES credentials are explicitly out of scope — they have dedicated rotation Lambdas that would conflict with Doppler-managed values. Redis is excluded too: it doesn't have a rotation Lambda yet, but adding one (matching the DB/SES pattern) is planned as follow-up work, which will keep it out of Doppler's management for the same reason DB/SES are excluded.

#### 🔗 Links to related PRs
None

#### ✅ Completion tasks

- [x] Added relevant tests
  - N/A — infra/tofu config change, no application code
- [x] Configuration changes:
  - [x] If new environment secrets are added, update in Tofu and set in AWS Secret Manager
    - Done via this PR + Doppler config, described above


noxmwalsh:
Do we already have these secrets entered into Doppler?

spokenbird:
> Do we already have these secrets entered into Doppler?

We do!
https://dashboard.doppler.com/workplace/08430c37e2a2889dc220/projects/safety-net-sebt-self-service-portal

.github/dependabot.yml
```diff
@@ -34,3 +34,16 @@ updates:
       interval: "weekly"
     cooldown:
       default-days: 7
+
+  - package-ecosystem: "terraform"
```
noxmwalsh, line 38: 👍🏻

tofu/config/dev-co/main.tf
```diff
@@ -56,27 +56,70 @@ data "aws_route53_zone" "main" {
...
+      description     = "JWT signing secret for the SEBT Portal API."
       recovery_window = 7
     }
-    "oidc" = {
-      description     = "MyColorado OIDC credentials for authentication."
+    "identifier_hasher_secret_key" = {
+      description     = "Identifier hashing secret for the SEBT Portal API."
+      recovery_window = 7
+    }
+    "cbms_client_id" = {
```
noxmwalsh, line 77: This is probably further out, but I'm thinking that it might be getting closer to time to disambiguate state-specific naming (like CBMS) in favor of something more generic (api_client_id).  This is something I think James' work in refactoring the contract/API spikes

tofu/config/dev-dc/main.tf
```diff
@@ -56,27 +56,58 @@ data "aws_route53_zone" "main" {
   name = "dc.sebt-portal.codeforamerica.app"
 }
 
-# Store DC-specific secrets in Secrets Manager. Each block represents a
-# separate set of secrets for a specific service or integration.
+# Store DC-specific secrets in Secrets Manager. Each key represents a
+# separate secret for a specific service or integration.
 module "state_secrets" {
-  source = "github.com/codeforamerica/tofu-modules-aws-secrets?ref=2.0.0"
+  source = "github.com/codeforamerica/tofu-modules-aws-secrets?ref=1880642d0546106d0c1f568304c0326b32b8cdbb" # 2.1.1
```
noxmwalsh, line 62: Maybe this isn't a big deal since I don't see anything weird in the Tofu Plan output, but the shared module [does have](https://github.com/codeforamerica/tofu-modules-aws-secrets/blob/main/main.tf) a reference to ignoring secret changes (via `ignore_secret_changes`) by default.  I'm not totally sure if this going to be an issue; I can't find where that variable is used in that module after giving it a `grep`, so it might actually be doing nothing.

tofu/config/dev-dc/main.tf
```diff
@@ -56,27 +56,58 @@ data "aws_route53_zone" "main" {
...
-    "socure" = {
-      description     = "Socure API credentials for identity verification."
+    "jwt_secret_key" = {
+      description     = "JWT signing secret for the SEBT Portal API."
       recovery_window = 7
     }
-    "pii_encryption" = {
-      description     = "PII encryption key ring for the SEBT Portal API (AES-256-GCM)."
+    "identifier_hasher_secret_key" = {
+      description     = "Identifier hashing secret for the SEBT Portal API."
```
noxmwalsh, line 74: I think this is used for hashing the `emailHash` column, so we should make sure this is the exact same key  to prevent misses and having to recreate users.

noxmwalsh (approved):
Approved with comments

tofu/config/dev-dc/main.tf
spokenbird: My understanding is ignore_secret_changes prevents terraform from treating external changes to secrets values as altering terraform state: https://registry.terraform.io/modules/terraform-aws-modules/secrets-manager/aws/latest#:~:text=no-,ignore_secret_changes,-Determines%20whether%20or

This is necessary for values that have automatic rotation through lambdas for example, or also for our case I believe because Doppler will ultimately set the values.

PR #524, opened 2026-07-28
DC-666(Bug): Restrict reseed endpoint to local dev/CI

noxmwalsh:
#### 🔗 Jira ticket
[DC-666](https://codeforamerica.atlassian.net/browse/DC-666)

#### ✍️ Description
Pen testing found POST /api/dev/seed/reseed/{scenarioName} callable without auth on the public DC lower environment. It was gated only by `ASPNETCORE_ENVIRONMENT=Development`, which is also how that host runs.

This change gates the endpoint on `Seeding:EnableDevEndpoints` (with the default value being `false`). The flag is set only in local launch profiles and Docker Compose. CI picks it up via the same launch profile. Deployed hosts (including public lower environments) leave it unset, so the action should return a 404 if there's an attempt to hit. 

The controller is also excluded from OpenAPI so it does not appear in Swagger when Development hosts still expose the UI.

Local and CI full-stack E2E reseed behavior should be unchanged and pass as before

#### 🔗 Links to related PRs
N/A

#### ✅ Completion tasks
After deploy to dev.dc, confirm an unauthenticated POST to /api/dev/seed/reseed/{scenarioName} returns 404, and that the route does not appear in Swagger.  That's all!

- [x] Added relevant tests
- [ ] Configuration changes:

[DC-666]: https://codeforamerica.atlassian.net/browse/DC-666?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


apps/portal/src/SEBT.Portal.Api/Controllers/Diagnostics/DevSeedController.cs
```diff
@@ -1,20 +1,23 @@
...
 /// </summary>
 [ApiController]
 [Route("api/dev/seed")]
 [AllowAnonymous]
-[Tags("Diagnostics")]
+[ApiExplorerSettings(IgnoreApi = true)]
 public class DevSeedController(
     IDatabaseSeeder databaseSeeder,
-    IWebHostEnvironment environment,
+    IOptions<SeedingSettings> seedingSettings,
```
necampanini: probably not a big deal but `IOptions<T>` wouldn't see a live AppConfig change and would require a process restart. 

maybe we can use an `IOptionsSnapshot<SeedSettings>` that the `TestErrorController` already uses?

necampanini (changes requested):
looks good! just the `IOptions` recommendation as a requested change

apps/portal/src/SEBT.Portal.Api/Controllers/Diagnostics/DevSeedController.cs
noxmwalsh: That's a good catch.  Addressed in f67bf26bd77d767653b5d44a5fb60afdfc5fda35

necampanini (approved):

PR #525, opened 2026-07-28
DC-596 Fix header logo alt text not announced by screen readers

spokenbird:
#### 🔗 Jira ticket
https://codeforamerica.atlassian.net/browse/DC-596

#### ✍️ Description
The header logo `Link` had both `aria-label="Home"` and an `<Image alt={t('bannerImageAltText')}>` child. Per the accessible-name computation algorithm ([accname 1.2](https://www.w3.org/TR/accname-1.2/#:~:text=three%20possible%20values%3A-,author,-name%20is%20generated) §4.3.2), `aria-label` on the link is resolved before the algorithm ever considers a descendant image's `alt` — so the `bannerImageAltText` content (e.g. "Logo: Colorado Deptartment of Human Services") was never announced by screen readers, only "Home."

It was also discovered that clicking the link would return the user to the login screen even if they were already authenticated to the dashboard. In the end it was decided after speaking with Ellie that we should just remove the link for now which this ticket does. 

Copy was also updated as there was a typo in the alt text. 1

#### 🔗 Links to related PRs
N/A

#### ✅ Completion tasks
- [X] Added relevant tests
- [X] Meets acceptance criteria in ticket
- [N/A] Configuration changes


necampanini (approved):
looks good! I saw some keys removed for Colorado but it made sense since they don't have the same `off-boarding` flow as DC

PR #534, opened 2026-08-03
feat(DC-568): state-backend config model (vocabulary records)

jamesmblair:
#### 🔗 Jira ticket

#### ✍️ Description
**Stack 1, PR 1 of 11.** We're adding a config-driven state-backend adapter. This stack builds it; nothing calls it yet, so runtime behavior doesn't change. Rationale: ADR-0020 (#544).

This PR is the config model — the C# records a state's YAML maps onto.

- `Core/StateBackends/Configuration/`: auth schemes, the five operation configs, field mapping, enum tables, disaggregation, caseId composition, result classifier, request binding.

**Focus:** does each config type explain itself from its name and one-line doc? It's the vocabulary a state author writes against.

#### ✅ Completion tasks

### DIFF
--- a/apps/portal/src/SEBT.Portal.Core/StateBackends/Configuration/Operations/FieldSources.cs
+++ b/apps/portal/src/SEBT.Portal.Core/StateBackends/Configuration/Operations/FieldSources.cs
@@ -1,0 +1,40 @@
+using System.Collections;
+
+namespace SEBT.Portal.Core.StateBackends.Configuration.Operations;
+
+/// <summary>One-or-more source property names for <see cref="FieldMapping.From"/>; hydrates from scalar or sequence YAML.</summary>
+public sealed class FieldSources : IEnumerable<string>
+{
+    private readonly IReadOnlyList<string> _sources;
+
+    public FieldSources(IReadOnlyList<string> sources)
+    {
+        ArgumentNullException.ThrowIfNull(sources);
+        if (sources.Count == 0)
+        {
+            throw new ArgumentException("A field mapping must name at least one source.", nameof(sources));
+        }
+
+        _sources = sources;
+    }
+
+    /// <summary>All source property names, in declared order.</summary>
+    public IReadOnlyList<string> All => _sources;
+
+    /// <summary>The single source; throws when the mapping names more than one.</summary>
+    public string Single => _sources.Count == 1
+        ? _sources[0]
+        : throw new InvalidOperationException(
+            $"Field mapping names {_sources.Count} sources; expected exactly one.");
+
+    public static implicit operator FieldSources(string source) => new(new[] { source });
+
+    public static implicit operator FieldSources(string[] sources) => new(sources);
+
+    // Null-tolerant so a null-conditional access (e.g. `mapping?.From`) converts without a warning.
+    public static implicit operator string?(FieldSources? sources) => sources?.Single;
+
+    public IEnumerator<string> GetEnumerator() => _sources.GetEnumerator();
+
+    IEnumerator IEnumerable.GetEnumerator() => GetEnumerator();
+}
--- a/apps/portal/src/SEBT.Portal.Core/StateBackends/Configuration/Operations/KeywordRules.cs
+++ b/apps/portal/src/SEBT.Portal.Core/StateBackends/Configuration/Operations/KeywordRules.cs
@@ -1,0 +1,14 @@
+namespace SEBT.Portal.Core.StateBackends.Configuration.Operations;
+
+/// <summary>Classifies an enum-typed field by keyword-substring scan of its <see cref="FieldMapping.From"/> source(s).</summary>
+public sealed record KeywordRules
+{
+    /// <summary>Canonical enum values in evaluation order; first-match-wins, so order is load-bearing.</summary>
+    public required List<string> Order { get; init; }
+
+    /// <summary>OUR canonical enum value → the substrings that indicate it.</summary>
+    public required Dictionary<string, List<string>> Map { get; init; }
+
+    /// <summary>Canonical enum value used when no keyword matches any source.</summary>
+    public required string Default { get; init; }
+}
--- a/apps/portal/src/SEBT.Portal.Core/StateBackends/Configuration/StateBackendCapabilities.cs
+++ b/apps/portal/src/SEBT.Portal.Core/StateBackends/Configuration/StateBackendCapabilities.cs
@@ -1,0 +1,12 @@
+namespace SEBT.Portal.Core.StateBackends.Configuration;
+
+public enum CardReplacementCapability
+{
+    None,
+    PerCase
+}
+
+public sealed record StateBackendCapabilities(
+    CardReplacementCapability CardReplacement,
+    bool AddressUpdate,
+    bool EnrollmentCheck);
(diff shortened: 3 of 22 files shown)


apps/portal/src/SEBT.Portal.Core/StateBackends/Configuration/Operations/KeywordRules.cs
```diff
@@ -0,0 +1,14 @@
+namespace SEBT.Portal.Core.StateBackends.Configuration.Operations;
+
+/// <summary>Classifies an enum-typed field by keyword-substring scan of its <see cref="FieldMapping.From"/> source(s).</summary>
+public sealed record KeywordRules
```
jamesmblair, line 4: Maybe revisit/harden this abstraction

apps/portal/src/SEBT.Portal.Core/StateBackends/Configuration/StateBackendConfiguration.cs
```diff
@@ -0,0 +1,24 @@
+using SEBT.Portal.Core.StateBackends.Configuration.Auth;
+using SEBT.Portal.Core.StateBackends.Configuration.Operations;
+
+namespace SEBT.Portal.Core.StateBackends.Configuration;
+
+public record StateBackendConfiguration
+{
+    public required Uri BaseUrl { get; init; }
```
noxmwalsh, line 8: I think `Core/StateBackends/Configuration/` might be kinda-sorta violating what ADR-0002 is stating as a rule:

> Kernel, Core, UseCases must not carry web/HTTP concept.

Whether or not we want to keep it as knowing 'transport' related concepts in `Core` is one thing, but we do have existing examples of this already in Core (with Smarty and Socure), so it's not like we don't have precedent.  

I'm just wondering on thoughts about it either being moved into `Infrastructure` or it's own adapter project.  I'm aware that if we want to do this, we'd probably want to do the same with the other examples I listed, so we can also just do this as a post-PR follow-up.

apps/portal/src/SEBT.Portal.Core/StateBackends/Configuration/StateBackendCapabilities.cs
```diff
@@ -0,0 +1,12 @@
+namespace SEBT.Portal.Core.StateBackends.Configuration;
```
noxmwalsh, line 1: Right now, listing an operation in the YAML is enough for the portal to treat it as a supported feature. That’s true even if the operation isn’t actually filled in yet.

The [ CO sample ](apps/portal/test/SEBT.Portal.Tests/Unit/Infrastructure/StateBackends/ConfigSamples/co.sample.yaml) from that PR does this for card replacement: it names the endpoint, but not how to build the request or tell success from failure. The portal would still show card replacement as available, and the first real attempt would fail as as the required parts not fully defined yet.

Do we want 'it’s in the config' to mean that users can do this, and if so should incomplete operations be rejected up front?  Should we at least have a warning?

apps/portal/src/SEBT.Portal.Core/StateBackends/Configuration/StateBackendCapabilities.cs
```diff
@@ -0,0 +1,12 @@
...
+
+public enum CardReplacementCapability
+{
+    None,
+    PerCase
```
noxmwalsh, line 6: Technically, we're doing a batch update in `ColoradoCardReplacementServices` in the MEF version.  Wouldn't this actually be a `Batch` type of enum here?

apps/portal/src/SEBT.Portal.Core/StateBackends/Configuration/Operations/KeywordRules.cs
noxmwalsh: Yeah, I think we need to make sure we're ok with what this is going to do.  After going through this, I understand now why we need this:

In `DcSummerEbtCaseService.cs`, we have:
```
if (householdType.Contains("OSSE") || eligibilityType.Contains("NSLP"))
    return IssuanceType.SummerEbt;
if (eligibilityType.Contains("FOOD") || eligibilityType.Contains("SNAP"))
    return IssuanceType.SnapEbtCard;
```

And I see this from one of the stacks

```
issuanceType:
  from: [HouseholdType, EligibilityType]
  keywordRules:
    order: [SummerEbt, SnapEbtCard, TanfEbtCard]
    map:
      SummerEbt:   [OSSE, NSLP]
      SnapEbtCard: [FOOD, SNAP]
      TanfEbtCard: [CASH, TANF]
```

So I see the goal here is to replicate what's going on here.  I think for what the adapter as it is now is trying to solve, it WOULD work DC as it is now.  However, as a method for other states to be able to use, I'm concerned that:

1. `Contains` is a case-insensitive check 
2. `Contains` immediately stops and returns once it finds the first match (unless it reaches the end of the string without one)

I do personally think there would be more flexibility if it was able to just take a Regex (because it can handle that case, plus future instances where we need case sensitivity, multiple matches, greeding etc.), but that would undermine the goal of having no expression language (since Regex itself is expressive).

apps/portal/src/SEBT.Portal.Core/StateBackends/Configuration/StateBackendConfiguration.cs
jamesmblair: Yeah, I think that's right. I think it makes sense to put it in the `Infrastructure.StateBackends` project. These types are only consumed by types in that project.

I'd prefer deferring to a follow-up PR though to avoid too much mess.

apps/portal/src/SEBT.Portal.Core/StateBackends/Configuration/Operations/FieldSources.cs
```diff
@@ -0,0 +1,40 @@
...
+        ? _sources[0]
+        : throw new InvalidOperationException(
+            $"Field mapping names {_sources.Count} sources; expected exactly one.");
+
+    public static implicit operator FieldSources(string source) => new(new[] { source });
+
+    public static implicit operator FieldSources(string[] sources) => new(sources);
+
+    // Null-tolerant so a null-conditional access (e.g. `mapping?.From`) converts without a warning.
+    public static implicit operator string?(FieldSources? sources) => sources?.Single;
```
spokenbird: Nit: `from:` accepts a sequence only alongside `keywordRules` (CONFIG-AUTHORING Step 5), but nothing enforces it. ValidateFieldMappings ([#537 StateBackendResponseMapper line 83](https://github.com/codeforamerica/sebt-self-service-portal/pull/537/changes#diff-d3f1e674c23568ac1c3540a949566f9e6081a27365f81733fb33c8829a5cb1e2R83)) doesn't check it, so a sequence without keywordRules loads clean but throws on first use when `From.Single` ([#537 line 305](https://github.com/codeforamerica/sebt-self-service-portal/pull/537/changes#diff-d3f1e674c23568ac1c3540a949566f9e6081a27365f81733fb33c8829a5cb1e2R305)) is called. 

The validation section of CONFIG-AUTHORING promises bad config fails at startup rather than on someone's first request, so this feels like it belongs there. Could it be a third check in that same loop? The other two put the field name in their errors so I think it's worth doing here too, since right now you just get "names 2 sources" with no idea which mapping it came from.

apps/portal/src/SEBT.Portal.Core/StateBackends/Configuration/StateBackendConfiguration.cs
spokenbird: Just want to make sure this is tracked somewhere? Didn't see the update in any of the rest of the PRs in the stack.

PR #535, opened 2026-08-03
feat(DC-568): per-operation Core ports + opaque caseId codec

jamesmblair:
#### 🔗 Jira ticket

#### ✍️ Description
**Stack 1, PR 2 of 11.** Still nothing calls the adapter.

Adds the Core ports — the interfaces the adapter will implement, one per operation.

- `IHouseholdLookupBackend`, `ICardReplacementBackend`, `IAddressUpdateBackend`, `IEnrollmentCheckBackend`, `IStateBackendHealth`, plus their request/result records and `WriteResult`.
- The opaque `caseId` codec: `OpaqueCaseId` and `CaseIdContext` — routing data travels inside an encoded caseId.

**Focus:** the port surface a caller depends on, and the opaque-token idea.

#### ✅ Completion tasks
- [ ] Added relevant tests

### DIFF
--- a/apps/portal/src/SEBT.Portal.Core/StateBackends/IAddressUpdateBackend.cs
+++ b/apps/portal/src/SEBT.Portal.Core/StateBackends/IAddressUpdateBackend.cs
@@ -1,0 +1,26 @@
+namespace SEBT.Portal.Core.StateBackends;
+
+public interface IAddressUpdateBackend
+{
+    Task<WriteResult> UpdateAddressAsync(
+        AddressUpdateRequest request, CancellationToken cancellationToken = default);
+}
+
+/// <summary>
+/// A household-routed mailing-address update: <see cref="CaseIds"/> MAY BE EMPTY (a zero-case
+/// household still updates), and a single backend call yields a single <see cref="WriteResult"/>.
+/// </summary>
+public sealed record AddressUpdateRequest(
+    string HouseholdIdentifier,
+    IReadOnlyList<string> CaseIds,
+    AddressUpdateAddress Address);
+
+/// <summary>The validated mailing-address scalars to persist.</summary>
+public sealed record AddressUpdateAddress
+{
+    public string? Line1 { get; init; }
+    public string? Line2 { get; init; }
+    public string? City { get; init; }
+    public string? State { get; init; }
+    public string? Zip { get; init; }
+}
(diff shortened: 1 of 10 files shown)


apps/portal/src/SEBT.Portal.Core/StateBackends/OpaqueCaseId.cs
```diff
@@ -0,0 +1,54 @@
...
+using System.Text.Json;
+
+namespace SEBT.Portal.Core.StateBackends;
+
+/// <summary>
+/// Packs a case's routing fields into a URL-safe base64 JSON token; reads compose it, writes decode
+/// it. Fixed here so both integration paths share one token namespace.
+/// </summary>
+/// <remarks>
+/// Opaque, not encrypted — it carries only routing identifiers the backend already returned.
```
noxmwalsh, line 11: The `dc.sample.yml` [file](https://github.com/codeforamerica/sebt-self-service-portal/pull/538/changes#diff-41759644078d74868313c6dc52a0ce1bd03a8d2e4cefa68676c30a9963f5a4dc) implies that:

```
fromContext:
  householdEmail: householdIdentifier
```

Would effectively be passing PII that would be opaque in base64 first, but could be reversed.  Given that a householdIdentifier could be any kind of Pii-based key (for example, SSN or TID; in the current DC case, it would be the user's email).

apps/portal/src/SEBT.Portal.Core/StateBackends/IAddressUpdateBackend.cs
```diff
@@ -0,0 +1,26 @@
...
+
+public interface IAddressUpdateBackend
+{
+    Task<WriteResult> UpdateAddressAsync(
+        AddressUpdateRequest request, CancellationToken cancellationToken = default);
+}
+
+/// <summary>
+/// A household-routed mailing-address update: <see cref="CaseIds"/> MAY BE EMPTY (a zero-case
+/// household still updates), and a single backend call yields a single <see cref="WriteResult"/>.
```
noxmwalsh, line 11: This says CaseIds may be empty and a zero-case household still updates, with routing via `HouseholdIdentifier`.  That part makes sense since DC does this today.

Later in the stack the driver never uses `HouseholdIdentifier` and the binder actually throws if there are zero decoded caseIds. That DC YAML also pulls household email from that base64 opaque token rather than the request envelope, which pushes PII into client-visible caseIds (as noted here: https://github.com/codeforamerica/sebt-self-service-portal/pull/535/changes#r3779957505).

So, to be clear, do we want to honor this port, or drop “MAY BE EMPTY” and document token-only routing?

PR #537, opened 2026-08-03
feat(DC-568): config-driven mapping engine

jamesmblair:
#### 🔗 Jira ticket

#### ✍️ Description
**Stack 1, PR 4 of 11.** Still nothing calls the adapter.

Adds the transform layer — the code that turns config into a response read, a request build, and a result verdict.

- `StateBackendResponseMapper`, `StateBackendRequestBinder`, `WriteResultClassifier`.
- Enrollment helpers: `EnrollmentRequestBuilder`, `ResponseCorrelator`, `CandidateExpander`, `OperationValidator`.
- No standalone tests here — exercised by the loader and driver PRs above it.

**Focus:** how config drives each transform, and the closed set of named primitives. There's no expression language — every transform is a named operation config picks from.

#### ✅ Completion tasks

### DIFF
--- a/apps/portal/src/SEBT.Portal.Infrastructure.StateBackends/Mapping/StateBackendResponseMapper.cs
+++ b/apps/portal/src/SEBT.Portal.Infrastructure.StateBackends/Mapping/StateBackendResponseMapper.cs
@@ -1,0 +1,631 @@
+using System.Globalization;
+using System.Text.Json;
+using SEBT.Portal.Core.Models.Household;
+using SEBT.Portal.Core.StateBackends;
+using SEBT.Portal.Core.StateBackends.Configuration;
+using SEBT.Portal.Core.StateBackends.Configuration.Operations;
+
+namespace SEBT.Portal.Infrastructure.StateBackends.Mapping;
+
+/// <summary>
+/// Maps a backend's raw JSON response into canonical domain types, driven by the operation's
+/// <see cref="StateBackendResponseMapping"/> and named enum tables.
+/// </summary>
+internal static class StateBackendResponseMapper
+{
+    /// <summary>The closed set of canonical field targets; a new canonical field means a new entry here — never reflection.</summary>
+    private static readonly IReadOnlyDictionary<string, FieldTarget> FieldTargets =
+        new Dictionary<string, FieldTarget>(StringComparer.Ordinal)
+        {
+            ["summerEBTCaseID"] = FieldTarget.String((c, v) => c.SummerEBTCaseID = v),
+            ["childFirstName"] = FieldTarget.String((c, v) => c.ChildFirstName = v),
+            ["childLastName"] = FieldTarget.String((c, v) => c.ChildLastName = v),
+            ["applicationId"] = FieldTarget.String((c, v) => c.ApplicationId = v),
+            ["ebtCardIssueDate"] = FieldTarget.DateTime((c, v) => c.EbtCardIssueDate = v),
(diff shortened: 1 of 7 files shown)


noxmwalsh:
I'm know it was mentioned in this section here regarding the unit test coverage in the description:

> No standalone tests here — exercised by the loader and driver PRs above it.

That’s a reasonable way to split the stack, but it means we don’t prove the mapping rules themselves here. There's some possible error cases (“wrong type of field”, “this keyword matched the wrong status", or “we treated an incomplete rule as success”, etc.) that we'd would probably want to test in a more isolated suite of unit tests.

I’d treat the later PRs as helping with the required coverage for this code, but I still think there's some value in adding focused tests on the mapping itself.

apps/portal/src/SEBT.Portal.Infrastructure.StateBackends/Mapping/StateBackendResponseMapper.cs
```diff
@@ -0,0 +1,631 @@
...
+    }
+
+    // Packs record + context routing fields into an opaque caseId token. A missing value packs
+    // empty; a later write that needs it fails loud.
+    private static string ComposeCaseId(JsonElement record, CaseIdComposition composition, CaseIdContext context)
+    {
+        var fields = new Dictionary<string, string>(StringComparer.Ordinal);
+        foreach ((string routingName, string sourceProperty) in composition.Fields)
+        {
+            fields[routingName] = JsonRead.AsString(record, sourceProperty) ?? string.Empty;
```
noxmwalsh, line 265: A lot of APIs send IDs and codes as numbers (for clarity, I'm referring to 0 vs. "0" for what's considered a valid JSON number).  The same response could pack the id into the case token fine, then error out when we map it as a normal field. 

I'm not sure if this is an issue with current expectation that'll apply when integrating with future partners, but it might be something we need to revisit later if it's an issue

apps/portal/src/SEBT.Portal.Infrastructure.StateBackends/Mapping/StateBackendResponseMapper.cs
```diff
@@ -0,0 +1,631 @@
...
+    }
+
+    private static bool IsApplicationBased(JsonElement record, StateBackendDisaggregation disaggregation)
+    {
+        string? discriminator = JsonRead.AsString(record, disaggregation.DiscriminatorField);
+
+        return disaggregation.Rule switch
+        {
+            DisaggregationRule.Presence => !string.IsNullOrEmpty(discriminator),
+            DisaggregationRule.ValueInSet =>
```
noxmwalsh, line 238: For Colorado we determine whether this row is an application by checking a field (which is `eligSrc`) and seeing if it's set with 'CBMS', 'PK', or whatever is currently set within that list. If that list is missing from the YAML, this code will keep going as normal and treat every row as not being an application.

In `CbmsResponseMapper.cs` 147-159
```
var applicationRows = students
    .Where(s => EligibilitySourceClassifier.IsApplicationBased(s.EligSrc))
    .Where(s => s.SebtAppId != null)
    .GroupBy(s => s.SebtAppId!);
```

In `EligibilitySourceClassifier.cs` 14-15
```
private static readonly HashSet<string> ApplicationSources =
        new(StringComparer.OrdinalIgnoreCase) { "CBMS", "PK" };
```

The CO sample includes the list, so the happy path is fine. Can we reject `valueInSet` here at load when the list is missing?

PR #539, opened 2026-08-03
feat(DC-568): API-key + OAuth client-credentials auth handlers

jamesmblair:
#### 🔗 Jira ticket

#### ✍️ Description
**Stack 1, PR 6 of 11.** Still nothing calls the adapter.

Adds the two auth schemes a state config can pick, plus secret resolution.

- `StateBackendApiKeyAuthHandler`, `StateBackendOAuthClientCredentialsAuthHandler`, `ConfigurationStateBackendSecretResolver`.

**Focus:** secrets by reference: config names a key, never the value. Check the OAuth handler's token caching.

#### ✅ Completion tasks
- [ ] Added relevant tests

### DIFF
--- a/apps/portal/src/SEBT.Portal.Infrastructure.StateBackends/Auth/StateBackendOAuthClientCredentialsAuthHandler.cs
+++ b/apps/portal/src/SEBT.Portal.Infrastructure.StateBackends/Auth/StateBackendOAuthClientCredentialsAuthHandler.cs
@@ -1,0 +1,125 @@
+using System.Net.Http.Headers;
+using System.Text.Json;
+using System.Text.Json.Serialization;
+using SEBT.Portal.Core.StateBackends;
+using SEBT.Portal.Core.StateBackends.Configuration.Auth;
+
+namespace SEBT.Portal.Infrastructure.StateBackends.Auth;
+
+/// <summary>
+/// Applies an OAuth2 client-credentials scheme: fetches a token, caches it until near expiry, and
+/// attaches it as a bearer token. The client secret is resolved via
+/// <see cref="IStateBackendSecretResolver"/> at token-fetch time — never inlined.
+/// </summary>
+public sealed class StateBackendOAuthClientCredentialsAuthHandler : DelegatingHandler
+{
+    // Refresh before actual expiry to avoid a token lapsing in flight.
+    private static readonly TimeSpan ExpiryLeeway = TimeSpan.FromSeconds(30);
+
+    private readonly StateBackendOAuthClientCredentialsAuthScheme _scheme;
+    private readonly IStateBackendSecretResolver _secretResolver;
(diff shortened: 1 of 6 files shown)


apps/portal/src/SEBT.Portal.Infrastructure.StateBackends/Auth/StateBackendOAuthClientCredentialsAuthHandler.cs
```diff
@@ -0,0 +1,125 @@
...
+            .ConfigureAwait(false);
+
+        return tokenResponse
+            ?? throw new InvalidOperationException("Token endpoint returned an empty response.");
+    }
+
+    private sealed record TokenResponse
+    {
+        [JsonPropertyName("access_token")]
+        public string AccessToken { get; init; } = string.Empty;
```
noxmwalsh, line 120: This would probably be fine if the API we're integrating with has the possibility of not requiring an access token, but I think we need to be careful here since it'll default into an empty string.  The previous check is just verifying we're getting a JSON response that's not empty

apps/portal/src/SEBT.Portal.Infrastructure.StateBackends/Auth/StateBackendOAuthClientCredentialsAuthHandler.cs
```diff
@@ -0,0 +1,125 @@
...
+
+        [JsonPropertyName("expires_in")]
+        public int ExpiresInSeconds { get; init; }
```
noxmwalsh, line 123: If the token response omits `expires_in` (or sends 0), this becomes `now - 30s`, so the cache is already expired.  I don't think the unit tests cover this scenario since they have a set `expires_in` and I'm not immediately sure what the full. consequences would be, but it would at least mean the cache would always be considered busted and we'd be fetching a new token every time.

apps/portal/src/SEBT.Portal.Infrastructure.StateBackends/Auth/StateBackendOAuthClientCredentialsAuthHandler.cs
```diff
@@ -0,0 +1,125 @@
...
+        ArgumentNullException.ThrowIfNull(secretResolver);
+        ArgumentNullException.ThrowIfNull(tokenClient);
+
+        _scheme = scheme;
+        _secretResolver = secretResolver;
+        _tokenClient = tokenClient;
+        _timeProvider = timeProvider ?? TimeProvider.System;
+    }
+
+    protected override async Task<HttpResponseMessage> SendAsync(
```
noxmwalsh, line 44: Do we have anything to handle when the API we're interacting with (in our current usecase, it would be CO's CBMS) returns a 401?  We'd probably at least want to try getting a new token maybe.

PR #540, opened 2026-08-03
feat(DC-568): config-driven driver — household lookup + health

jamesmblair:
#### 🔗 Jira ticket

#### ✍️ Description
**Stack 1, PR 7 of 11.** Still nothing calls the adapter.

Adds the driver itself, `ConfigurableStateBackend`, starting with two read operations: household lookup and health.

- This is where config, primitives, and mapping come together against (mocked) HTTP.

**Focus:** the request→send→map pipeline for a read. The driver grows one operation per PR from here.

#### ✅ Completion tasks
- [ ] Added relevant tests

### DIFF
--- a/apps/portal/src/SEBT.Portal.Infrastructure.StateBackends/ConfigurableStateBackend.cs
+++ b/apps/portal/src/SEBT.Portal.Infrastructure.StateBackends/ConfigurableStateBackend.cs
@@ -1,0 +1,131 @@
+﻿using System.Text;
+using System.Text.Json;
+using System.Text.Json.Nodes;
+using SEBT.Portal.Core.Models.Household;
+using SEBT.Portal.Core.StateBackends;
+using SEBT.Portal.Core.StateBackends.Configuration;
+using SEBT.Portal.Core.StateBackends.Configuration.Operations;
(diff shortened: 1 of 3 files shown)


apps/portal/src/SEBT.Portal.Infrastructure.StateBackends/ConfigurableStateBackend.cs
```diff
@@ -0,0 +1,131 @@
...
+        {
+            using HttpResponseMessage response = await _httpClient
+                .SendAsync(request, cancellationToken)
+                .ConfigureAwait(false);
+
+            return new StateBackendHealth(response.IsSuccessStatusCode);
+        }
+        catch (HttpRequestException)
+        {
+            return new StateBackendHealth(IsHealthy: false);
```
noxmwalsh, line 61: Does this cover timeouts? In .NET, when `HttpClient` hits its timeout, that usually comes through as a cancellation exception, not as `HttpRequestException`. So if, for example, CBMS is slow to respond (which did used to be the case) rather than unreachable, this line would throw instead of returning unhealthy.

PR #559, opened 2026-08-05
[DC-159] Fetch a fresh OIDC token for each Doppler plan and apply step

spokenbird:
#### 🔗 Jira ticket
https://codeforamerica.atlassian.net/browse/DC-159

#### ✍️ Description
The apply after #558 merged failed again on both dev-dc and dev-co with the same `This OIDC token was used previously` error, even with a single shared provider instance.

Root cause (confirmed with James Armes): `tofu apply` without a saved plan file isn't one operation — it computes a fresh plan internally and then executes it, and the Doppler provider re-authenticates between those two phases. We were only fetching one OIDC token per job, so the first phase's exchange succeeded and burned the token before the second phase could use it.

Fix: split `deploy-dc`/`deploy-co` into an explicit `tofu plan -out=tfplan`, then a freshly-fetched OIDC token, then `tofu apply tfplan`. Applying a saved plan file only needs one provider authentication, so two separately-fetched tokens each satisfy exactly one exchange.

### DIFF
--- a/.github/workflows/deploy-ecr.yaml
+++ b/.github/workflows/deploy-ecr.yaml
@@ -295,12 +295,6 @@ jobs:
           docker push ${API_REPO}:${IMAGE_TAG}
           docker push ${WEB_REPO}:${IMAGE_TAG}
 
-      - name: Get an OIDC token for Doppler
-        run: |
-          DOPPLER_OIDC_TOKEN=$(curl -H "Authorization: bearer $ACTIONS_ID_TOKEN_REQUEST_TOKEN" \
-            "$ACTIONS_ID_TOKEN_REQUEST_URL&audience=https://github.com/$GITHUB_REPOSITORY_OWNER" | jq -r '.value')
-          echo "DOPPLER_OIDC_TOKEN=$DOPPLER_OIDC_TOKEN" >> $GITHUB_ENV
-
       - name: Set up Python
         uses: actions/setup-python@5fda3b95a4ea91299a34e894583c3862153e4b97 # v7
         with:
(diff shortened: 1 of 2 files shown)


.github/workflows/deploy-ecr.yaml
```diff
@@ -318,12 +312,32 @@ jobs:
...
-      - name: OpenTofu Apply
+      # Doppler OIDC tokens are single-use. Plan and apply each need their own
+      # provider authentication, so each gets its own freshly-fetched token —
+      # reusing one token for both fails with "This OIDC token was used
+      # previously".
+      - name: Get an OIDC token for Doppler (plan)
+        run: |
+          DOPPLER_OIDC_TOKEN=$(curl -H "Authorization: bearer $ACTIONS_ID_TOKEN_REQUEST_TOKEN" \
+            "$ACTIONS_ID_TOKEN_REQUEST_URL&audience=https://github.com/$GITHUB_REPOSITORY_OWNER" | jq -r '.value')
+          echo "DOPPLER_OIDC_TOKEN=$DOPPLER_OIDC_TOKEN" >> $GITHUB_ENV
```
noxmwalsh, line 323: This is OK due to the size and that it's a single-use token, but https://docs.github.com/en/actions/reference/security/secure-use (look under mask sensitive data) suggests masking secrets like this:

```
echo "::add-mask::$DOPPLER_OIDC_TOKEN"
echo "DOPPLER_OIDC_TOKEN=$DOPPLER_OIDC_TOKEN" >> "$GITHUB_ENV"
```

I think this echo might be outputted in the logs, so the masking will help cover that and future cases like this.

noxmwalsh (approved):
Approved with non-blocking comment (it's something we need to keep in mind for long-term secrets, but not necessarily for this case)

PR #563, opened 2026-08-06
[DC-356] NuGet lockfiles + locked-mode CI enforcement

jamesmblair:
#### 🔗 Jira ticket
https://codeforamerica.atlassian.net/browse/DC-356

#### ✍️ Description
Completes the lockfile half of DC-356 for the monorepo. (`packageSourceMapping` in the root `nuget.config` already existed — squatting defense was in place.)

- Enable lockfiles solution-wide via `RestorePackagesWithLockFile` in the root `Directory.Build.props` — every project pins exact direct + transitive versions.
- Commit the 14 generated `packages.lock.json` (one per `SEBT.slnx` project). **These are generated — the only hand-written changes are `Directory.Build.props` (+1 line) and `scripts/ci/build-backend.sh` (+7).**
- CI restore runs `--locked-mode` (CI-gated in `build-backend.sh`) — a missing or drifted lockfile fails the build. Local dev is unaffected, so devs can still add and refresh packages.

Verified: locked-mode restore passes clean against the committed lockfiles; tampering one content hash fails restore with `NU1403`.

#### 🔗 Links to related PRs
dc-connector `packageSourceMapping`: https://github.com/codeforamerica/sebt-self-service-portal-dc-connector/pull/54

#### ⚠️ Known follow-up
Dependabot doesn't refresh the lockfiles of *referencing* projects when it bumps a package in a referenced project ([dependabot-core#12318](https://github.com/dependabot/dependabot-core/issues/12318), open). So a core-lib bump (e.g. EF Core in `Infrastructure`) will fail locked-mode CI on `Api`/`Tests` until the lockfiles are regenerated with `dotnet restore --force-evaluate`. Auto-refresh mitigation to be tracked separately.

#### ✅ Completion tasks
- [ ] Added relevant tests — N/A: build configuration; verified via locked-mode restore + tamper test (see Description)
- [x] Configuration changes: none (no environment variables, secrets, or appsettings)


adbergen (changes requested):
Hey @jamesmblair ! Reviewed both DC-356 PRs. The connector one (https://github.com/codeforamerica/sebt-self-service-portal/pull/54) looks good and approved, though we should note on the ticket that the connector repo itself still has no lockfiles (the floating 0.0.4-* contract ref makes locked mode awkward, so worth recording the decision either way).

This portal one (https://github.com/codeforamerica/sebt-self-service-portal/pull/563) needs another pass before merge: the --locked-mode flag went into scripts/ci/build-backend.sh, but PR CI (state-ci + playwright) runs the duplicate copy at .github/workflows/scripts/build-backend.sh, which still restores unlocked, so drift merges green and would only blow up at DC release time. Also the ECR image build and the connector/publish steps never see the lockfiles at all, and since RestorePackagesWithLockFile is repo-wide now, unlocked restores silently rewrite stale lockfiles instead of failing.

Suggestion that fixes most of it in one line: drop the shell flag and put true in Directory.Build.props next to the lockfile property. That covers every restore including implicit ones.
Caveats: Docker needs CI as a build ARG + lockfiles COPYed in, and the nested dc-connector checkout needs handling.

adbergen (changes requested):
Thanks for addressing the requested changes! Found 2 more following recent changes, then should be good to go.

1. Noticed Build DC IIS bundle fails. Seems like a real issue and not just flaky. The project's runtime identifiers changed across the whoel Api closure. Seems like DC Prod artifact. Solution could be declare `<RuntimeIdentifiers>win-x64</RuntimeIdentifiers>` on `SEBT.Portal.Api.csproj` (so every future `--force-evaluate` refresh keeps the win-x64 sections, not just this one) and regenerate the closure lockfiles.

2. Stale lockfiles so just need to refresh locklfiles since some merges came in yesterday.

adbergen (approved):

PR #564, opened 2026-08-06
[DC-600] Canonicalize phone identifiers to E.164 before hashing

jamesmblair:
#### 🔗 Jira ticket
https://codeforamerica.atlassian.net/browse/DC-600

#### ✍️ Description
The resolver only trimmed phone household identifiers (Colorado) before hashing them as `HouseholdIdentifierHash`. The same number arrived in different formats — a JWT `phone` claim in E.164, `user.Phone`, parens, dots, a bare country code — and each format hashed differently. So cooldown, cache, and analytics could split one person into several.

`HouseholdIdentifierResolver` now runs every phone source (dev override, JWT claim, user record) through `PhoneNormalizer` and canonicalizes to **E.164** (`+1NNNNNNNNNN`). If libphonenumber can't validate the value as a US number, it resolves to `null` and the resolver falls through to the next source instead of hashing a non-canonical value. For phone-only states with no valid source, that surfaces as `Unauthorized`.

**Two things reviewers should know:**

- **E.164, not the ticket's literal 10-digit.** The ~10k existing Colorado `CardReplacementRequests` rows were hashed from the E.164 the IdP emits. E.164 keeps those cooldowns intact; 10-digit national would silently reset every live cooldown. Any legacy rows in another format age out within the 14-day window — no backfill, since we don't store the original cleartext.
- **The ticket's framing is a misattribution.** Nothing computes `HouseholdIdentifierHash` from a CBMS response — it's the portal-resolved identifier. The CO connector already normalizes phone before every CBMS call and keys its cache on `phone10`, so caching and CBMS were never broken. The real bug is resolver-side format drift across requests.

The mock-repo change is dev-only: its lookup now strips a leading US country code so E.164 identifiers match the 10-digit seeded personas. It's a no-op for existing 10-digit inputs.

#### 🔗 Links to related PRs
None.

#### ✅ Completion tasks

- [x] Added relevant tests
- [X] Configuration changes: N/A

### DIFF
--- a/apps/portal/src/SEBT.Portal.Infrastructure/Repositories/MockHouseholdRepository.cs
+++ b/apps/portal/src/SEBT.Portal.Infrastructure/Repositories/MockHouseholdRepository.cs
@@ -1141,6 +1141,14 @@ public class MockHouseholdRepository : IHouseholdRepository
         }
 
         var digits = Regex.Replace(phone.Trim(), @"\D", "");
+
+        // Resolved identifiers now arrive in E.164 (+1NNNNNNNNNN); strip the leading
+        // US country code so they match the 10-digit keys the personas are seeded with.
+        if (digits.Length == 11 && digits[0] == '1')
+        {
+            digits = digits[1..];
+        }
+
         return string.IsNullOrEmpty(digits) ? null : digits;
     }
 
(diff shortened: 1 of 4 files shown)


noxmwalsh (approved):
I think this looks good!  Approved

PR #567, opened 2026-08-11
[DC-690] Fix API configuration precedence so environment variables override state appsettings

adbergen:
#### 🔗 Jira ticket
https://codeforamerica.atlassian.net/browse/DC-690

#### ✍️ Description
`WebApplication.CreateBuilder` registers the environment-variable and command-line providers before `Program.cs` appended `appsettings.{state}.json`, so the state overlay silently won over values operators set via env vars — contradicting twelve-factor expectations and the precedence comment in `Program.cs`.

- Extracted the overlay wiring into `StateOverlayConfiguration.AddStateOverlay`, which inserts the state JSON source just below the app-level (unprefixed) environment-variable provider instead of appending it. The prefixed host sources (`ASPNETCORE_`/`DOTNET_`) stay below the appsettings files.
- Effective precedence is now (later wins): base appsettings → user secrets → state overlay → env vars → command-line args → AWS AppConfig Agent.
- Added `StateOverlayConfigurationTests` (7 tests) built on the real `WebApplication.CreateBuilder` chain — the existing `ConfigurationProviderPriorityTests` hand-assemble their own builder, which is why this never surfaced. Each acceptance criterion has a dedicated test; the env-var and command-line tests were verified failing against the old ordering before the fix.
- Updated the `Program.cs` precedence comment to document the full chain.
- Behavior note: local Docker Compose env vars (`Seeding__*`, `Redis__*`, `SmtpClientSettings__*`) now correctly win over any local state overlay values, matching the ticket's intent.

Also in this branch (separate commit): replaced the `/conventions` Claude skill with a diff-scoped `/code-standards` skill covering both branch-diff and whole-path reviews.

#### 🔗 Links to related PRs
N/A

#### ✅ Completion tasks

- [x] Added relevant tests
- [ ] Configuration changes:
  - [ ] If new environment variables are added, update in Tofu — N/A, no new variables
  - [ ] If appsetttings are changed, update in AWS AppConfig — N/A, precedence-only change


adbergen:
# QA Summary — feature/DC-690-config-env-var-precedence

## What Changed

Fixed the order the API reads its configuration sources. Previously, values in a state's settings file (`appsettings.dc.json` / `appsettings.co.json`) silently overrode values operators set via environment variables. Now environment variables win over the state file, command-line arguments win over environment variables, and AWS AppConfig (when configured) still wins over everything — matching standard .NET/twelve-factor expectations. No user-facing screens changed; the risk surface is "the portal picks up different configuration values than before" in environments where the same setting is defined in both places.

## Affected Areas

- API startup configuration in every deployed environment (DC dev, CO dev, DC prod)
- Any setting defined in both an environment variable and the state settings file — previously the file won; now the environment variable wins
- Local Docker Compose development (compose-set variables like seeding and Redis settings now correctly take effect)

## Test Cases

### Configuration Precedence (ops-level, dev environment)

- [ ] 1. In a dev environment (or locally), pick a setting that exists in the state settings file, set a different value for it as an environment variable, and restart the API. The API should use the environment variable's value. Before this change it used the file's value.
- [ ] 2. Remove the environment variable and restart. The API should fall back to the state file's value (state file still overrides the base settings).
- [ ] 3. CO dev only: confirm AppConfig-managed settings and feature flags still apply — AppConfig should still override everything, including environment variables.

### Regression Checks (user-facing smoke)

- [ ] 4. DC: sign in, land on the dashboard, confirm household/case data loads and page copy renders normally.
- [ ] 5. CO: sign in via OIDC, confirm the dashboard loads and CO theming/content look normal.
- [ ] 6. Confirm feature-flag-driven UI (e.g., anything currently dark-launched) is in the same state as before the deploy.
- [ ] 7. Enrollment checker: run one eligible and one ineligible check, confirm results unchanged.

## Test Credentials

- DC seeded persona — `sebt.dc+co-loaded@codeforamerica.org` — OTP arrives in Mailpit locally (http://localhost:8025); in deployed dev, use the team's standard OTP mailbox
- CO seeded persona — `sebt.co+co-loaded@codeforamerica.org` — same OTP retrieval
- Other scenario suffixes exist (`verified`, `expired`, `review`, `dc-mixed`, `co-active`, …) if a specific dashboard state is needed

## Environment Notes

- The precedence change only shows up where a setting is defined in **both** an environment variable and the state settings file — audit any environment relying on the old (file-wins) behavior before deploy
- Local: `docker compose up -d` variables (seeding, Redis, SMTP) now correctly override local state-file values — if your local setup depended on the file winning, update your `.env`/compose overrides
- Test on both a DC and a CO environment, since each loads a different state overlay

apps/portal/src/SEBT.Portal.Api/Startup/StateOverlayConfiguration.cs
```diff
@@ -0,0 +1,47 @@
...
+        // contains prefixed host sources (ASPNETCORE_/DOTNET_) that sit before the
+        // appsettings files and must stay below the overlay.
+        var environmentVariablesIndex = configuration.Sources.ToList().FindIndex(
+            source => source is EnvironmentVariablesConfigurationSource { Prefix: null or "" });
+
+        if (environmentVariablesIndex >= 0)
+        {
+            configuration.Sources.Insert(environmentVariablesIndex, overlaySource);
+        }
+        else
```
necampanini: I don't think we can hit the `else` branch currently, but could we throw or log a warning if that ever isn't the case? mostly just to give us some hint in the future if configuration ordering quietly fell back to its previous order and we're wondering why

necampanini (changes requested):
looks good, just requesting the small change

necampanini (approved):

PR #583, opened 2026-08-12
Load vendor Urbanist and Atkinson Hyperlegible Next fonts locally to fix flaky CI builds

spokenbird:
#### 🔗 Jira ticket
https://codeforamerica.atlassian.net/browse/DC-723

#### ✍️ Description
CI's frontend build step has been intermittently failing with:

```
Module not found: Can't resolve '@vercel/turbopack-next/internal/font/google/font'
```

This is a [known Turbopack regression in Next.js 16.2.x](https://github.com/vercel/next.js/issues/91653) ([related](https://github.com/vercel/next.js/issues/92671)): `next/font/google` makes Turbopack's Rust HTTP client fetch the font binary from `fonts.gstatic.com` **at build time**, and that client intermittently fails to establish the connection. It's not a Google Fonts outage — the one merged upstream fix for this class of bug ([vercel/next.js#92713](https://github.com/vercel/next.js/pull/92713)) was scoped to Windows ARM64, so it doesn't cover our `ubuntu-latest` GitHub Actions runners, and I found no comprehensive fix in any release since.

Confirmed this isn't caused by any single PR's diff — it hit two unrelated, concurrently-running PRs at the same time:
- #577 (`dc - Build & Test`): https://github.com/codeforamerica/sebt-self-service-portal/actions/runs/31637200328/job/94250446864
- #575 (`Playwright Integration E2E (DC)` and `Pa11y Accessibility Tests`): https://github.com/codeforamerica/sebt-self-service-portal/actions/runs/31637290619/job/94250752810

Fix: load the two Google-Fonts-sourced typefaces (Urbanist for DC, Atkinson Hyperlegible Next for CO) via `next/font/local` instead, vendoring the `.woff2` files (OFL-licensed, freely redistributable) the same way Museo Slab already is. This removes the build-time network dependency entirely rather than papering over an upstream bug we don't control.

Both fonts are variable fonts, so a single vendored file per font covers the full weight range (`400 700`) — no per-weight files needed.

Scope:
- `packages/design-system/design/scripts/generate-fonts.js` — moved `urbanist` and `atkinson hyperlegible` from `GOOGLE_FONTS_MAP` to `LOCAL_FONTS_MAP`
- `packages/design-system/design/scripts/generate-fonts.test.js` — updated assertions to expect `next/font/local` for these two fonts
- Vendored `.woff2` files under `public/fonts/urbanist/` and `public/fonts/atkinson-hyperlegible-next/` in both `SEBT.Portal.Web` and `SEBT.EnrollmentChecker.Web`
- `.gitignore` carve-outs for the two new font directories, mirroring the existing `museo-slab` entry

Verified with a full production build (`next build`, Turbopack) for both DC and CO states in both apps — no font-fetch step occurs anymore, confirmed by the generated `fonts.ts` output using `localFont` instead of the Google Fonts loader.

#### 🔗 Links to related PRs
N/A

#### ✅ Completion tasks
- [x] Added relevant tests
- [x] Meets acceptance criteria in ticket — N/A, no ticket
- [ ] Configuration changes: N/A (no env vars, secrets, or appsettings involved)


adbergen:
Love killing the gstatic fetch, and the variable-font swap is a net win on bytes (one variable file + one preload vs. three static files). A few things surfaced in review — three I'd fix before merge, the rest can be follow-ups.

**Before merge:**

1. **Glyph coverage**: the old `next/font/google` path self-hosted *every* subset with per-subset `unicode-range` — the `subsets: ['latin']` option only controls preloading (verifiable in `next/dist/compiled/@next/font/dist/google/find-font-files-in-css.js`). So the latin-only vendored files are a real regression: a household name like "Nguyễn" or "Łukasz" now renders half Urbanist, half system font mid-word. Can we re-subset with latin + latin-ext in one file?
2. **OFL compliance + provenance**: Urbanist and Atkinson Hyperlegible Next are SIL OFL, and we now redistribute them publicly with no `OFL.txt`, no upstream version, and no record of the subsetting command — which also means nobody can reproduce or verify the subset when we need to fix #1. An `OFL.txt` plus a one-line source/subset note per font dir closes both (the museo-slab entry at least names Font Squirrel; these name nothing). They also bypass the license audit, which only scans npm metadata.
3. **`weight: '400 700'` clamps the variable axis** and contradicts the comment right above it ("one file covers the full weight range" — the files carry ~100–900 / 200–800). Any future 300/800 usage silently clamps, and USWDS's `b,strong { font-weight: bolder }` inside a 700 heading now clamps to 700. Declaring the real ranges costs nothing — the glyph data already ships.

**Cheap wins while you're in here:** the file header still says the generator "uses next/font/google" and the `LOCAL_FONTS_MAP` header still says "fonts that aren't on Google Fonts" (right above two Google fonts) — both now steer the next dev back onto the flaky path; the `variable:` fields in the map entries are never read (`buildFontLoader` takes the CSS var from its caller — same trap as the old `--font-urbanist` bug); an `existsSync` on `src[].path` at generation time would turn a missing binary into a clear error instead of a module-not-found inside a gitignored generated file; and the `fonts.gstatic.com` / `fonts.googleapis.com` CSP entries in `proxy.ts` are now provably dead — removing them makes any future remote-font regression a visible CSP violation.

**Follow-up ticket material** (happy to file): the google branch is still live and untested — a future state set to "Public Sans" (USWDS's default, still in `GOOGLE_FONTS_MAP` even though its woff2 already ships via the USWDS copy step) silently reintroduces this exact CI flake; `tokens:all` now generates an uncompilable `fonts.ts` (it runs the generator with cwd = packages/design-system, where `../public/fonts/` doesn't exist); the binaries are duplicated per app with nothing keeping them in sync (one source dir under `packages/design-system` would fix drift, the doubled image bytes, and the growing `.gitignore` negation lists); and `design/fonts.test.ts` — the contract test written for the last fonts.ts bug — never actually runs (vitest's `include` misses `design/`, and its SCSS fixture path is stale).

PS, unrelated to this PR but spotted in passing: `content/locales/am/dc/outage.json` has four Hebrew characters mid-Amharic — looks like a translation artifact worth flagging to the content team.

adbergen (changes requested):
See summary comment

noxmwalsh:
I just want to follow-up that @adbergen is correct in that we do need item 2 addressed at a minimum before we merge this in.

For Urbanist: https://github.com/coreyhu/Urbanist/blob/main/OFL.txt
For Atkinson: https://github.com/google/fonts/blob/main/ofl/atkinsonhyperlegiblenext/OFL.txt

> 2) Original or Modified Versions of the Font Software may be bundled,
> redistributed and/or sold with any software, provided that each copy
> contains the above copyright notice and this license. These can be
> included either as stand-alone text files, human-readable headers or
> in the appropriate machine-readable metadata fields within text or
> binary files as long as those fields can be easily viewed by the user.

We need to include the licenses (OFL.txt for both) if we want to use these fonts and stay compliant with license requirements for this project

adbergen (approved):
LGTM!

noxmwalsh (commented):
I think the licensing requirement is met now for OFL, so my main concern is addressed

PR #585, opened 2026-08-13
[DC-685] Ignore portal session cookie on anonymous enrollment checker API

adbergen:
#### 🔗 Jira ticket
https://codeforamerica.atlassian.net/browse/DC-685

#### ✍️ Description
Investigation of DC-685 found that `portal_user_id`/`portal_user_email`/`portal_user_phone` appear on enrollment-checker logs in CO prod because:

- CO prod serves the portal and the enrollment checker from the same origin (`.../login` and `/enrollment` on one host), and the portal session cookie is host-only with `Path=/`, so it rides along on the checker's same-origin `POST /api/enrollment/check`.
- `[AllowAnonymous]` skips authorization, not authentication: the JwtBearer cookie fallback in `Program.cs` still authenticates the request, and `PortalUserEnricher` then stamps portal identity onto every log event in it.

**Fix:** the session cookie is no longer read for the anonymous enrollment-checker API. `AuthCookies.AllowsCookieAuthentication` excludes `/api/enrollment/*` from the cookie fallback; the `Authorization` header path is unchanged. Side effect fixed for free: a portal user with a policy-violating (stale) session no longer gets their portal cookie cleared just by using the checker.

**Tests** (TDD, watched fail first):
- Unit: 8 path cases on the new predicate (enrollment paths refuse cookie auth; portal paths unchanged). `AuthCookiesTests` moved to `Unit/Api/Services/` per the mirrored-namespace convention.
- Integration: a valid-signature token without `auth_time` authenticates but fails the session-lifetime policy, which clears the cookie — so pre-fix, `GET /api/enrollment/features` with that cookie produced a `Set-Cookie` header proving authentication ran. The new test asserts no `Set-Cookie`; a control test proves the same token on `/api/auth/status` still 401s and clears the cookie.

Also in this branch: `SSH.NET` pinned to 2026.0.0 in the test project — a fresh advisory (GHSA-q939-rpr3-3284, transitive via Testcontainers) breaks every clean restore under our NuGetAudit-as-errors policy. Separate commit; happy to split it out if we want it merged faster.

#### 🔗 Links to related PRs
N/A

#### ✅ Completion tasks

- [x] Added relevant tests
- [ ] Configuration changes:
  - [ ] If appsetttings are changed, update in AWS AppConfig — N/A


apps/portal/test/SEBT.Portal.Tests/SEBT.Portal.Tests.csproj
```diff
@@ -32,6 +32,10 @@
         <PackageReference Include="OpenTelemetry.Exporter.InMemory" Version="1.15.3" />
         <PackageReference Include="RichardSzalay.MockHttp" Version="7.0.0" />
         <PackageReference Include="StackExchange.Redis" Version="2.13.17" />
+        <!-- Transitive dependency of Testcontainers; pinned directly so NuGetAudit
+             restores clean after GHSA-q939-rpr3-3284 (fixed in 2026.0.0). Remove the
+             pin once Testcontainers depends on a patched version. -->
+        <PackageReference Include="SSH.NET" Version="2026.0.0" />
```
noxmwalsh, line 38: I think we can just upgrade to 4.14.0 for `TestContainers` to address this.  We can leave that for another PR, but I think that should fix this particular issue.

noxmwalsh (approved):
Approved with comments.  There's an upgrade that'll address the transitive issue that I think we should consider, but I don't think that should block this ticket.

apps/portal/test/SEBT.Portal.Tests/SEBT.Portal.Tests.csproj
adbergen: Confirmed, 4.14.0 pulls SSH.NET 2026.0.0, so the pin becomes unnecessary.

PR #586, opened 2026-08-14
[DC-703] Route Api and UseCase dependencies through Core state connector ports

adbergen:
#### 🔗 Jira ticket
https://codeforamerica.atlassian.net/browse/DC-703

#### ✍️ Description
Routes the remaining Api and UseCases dependencies through Core ports, completing the pattern household reads already follow (`IHouseholdRepository` → Infrastructure adapter):

- **New `Core.StateConnector` boundary**: `IStateEnrollmentCheckService`, `IStateAddressUpdateService`, `IStateCardReplacementService` + field-exact boundary models/enums. Infrastructure adapters (`Infrastructure/StateConnector/`) map to the plugin contract with exhaustive enum switches that throw on unmapped members, so contract drift fails loudly at the boundary.
- **UseCases no longer reference `StatesPlugins.Interfaces` at all** — the conditional plugin-contract reference block is deleted from the csproj.
- **Api controllers have zero Infrastructure usings**: the outage port (`IOutagePageStateResolver`/`OutagePageState`) and the Socure enrichment parser move to `Core.Services` (both relocations are namespace-only diffs); `TestErrorController`'s inline `SmartyAddressUpdateService` construction is extracted behind a new `IAddressVerificationDiagnostics` Core port (semantic scenarios — no HTTP status codes in Core) implemented in Infrastructure.
- **Distinct naming**: Core's Smarty validation port is renamed `IAddressUpdateService` → `IAddressVerificationService` (impls `SmartyAddressVerificationService` / `PassThroughAddressVerificationService`), so it can no longer be confused with the state-connector write-back port. The plugin contract itself is untouched.
- Decision + model-duplication trade-off recorded in [docs/adr/0020-core-ports-for-state-connector-access.md](../blob/feature/DC-703-core-ports/docs/adr/0020-core-ports-for-state-connector-access.md).

Behavior is unchanged: handler and relocated-file diffs are type/using-only (verified line-by-line against main), existing handler/controller/parser tests pass with zero assertion changes, and 30 new adapter/diagnostics tests (TDD, watched fail first) cover every mapped field and enum member. Also aboard: the SSH.NET audit pin (same commit as other open PRs; merges no-op).

#### 🔗 Links to related PRs
N/A

#### ✅ Completion tasks

- [x] Added relevant tests
- [ ] Configuration changes:
  - [ ] If appsetttings are changed, update in AWS AppConfig — N/A


apps/portal/test/SEBT.Portal.Tests/Unit/Infrastructure/StateConnector/PluginEnrollmentCheckServiceTests.cs
```diff
@@ -0,0 +1,289 @@
...
+
+    public PluginEnrollmentCheckServiceTests()
+    {
+        _sut = new PluginEnrollmentCheckService(_plugin);
+    }
+
+    private static PluginEnrollmentCheckResult EmptyPluginResult() =>
+        new() { Results = [] };
+
+    public static TheoryData<PluginEnrollmentStatus> AllPluginEnrollmentStatuses()
```
noxmwalsh, line 32: This only walks the plugin list of statuses, so if someone adds a new value on the Core side and forgets the plugin copy, CI will likely still pass. 

The card-replacement test has the same gap in the other direction. We also already have `EnumParityTests` to confirm both copies of an enum stay in lockstep. 

Can we trying adding `EnrollmentStatus`, `EligibilityType`, and `CardReplacementReason` there at the `EnumParityTests`. These mapping tests can stay; they still have value in that they're checking the value translates correctly.

apps/portal/src/SEBT.Portal.Core/StateConnector/Address.cs
```diff
@@ -0,0 +1,13 @@
+namespace SEBT.Portal.Core.StateConnector;
+
+/// <summary>
+/// Represents an address on file for a household.
+/// </summary>
+public class Address
```
noxmwalsh: We already have a household address type with these same five fields. This adds a second copy in `Core`, so the `update-address` handler has to alias both and copy field by field before the adapter copies them again into the plugin type. 

Can `AddressUpdateRequest` use `Core.Models.Household.Address` instead, and leave the plugin mapping in the adapter?  WDYT?

apps/portal/src/SEBT.Portal.Infrastructure/Services/IAddressVerificationDiagnostics.cs
```diff
@@ -0,0 +1,35 @@
...
+/// exercise the verification service's error-handling and logging paths without any network
+/// activity.
+/// </summary>
+public interface IAddressVerificationDiagnostics
+{
+    /// <summary>
+    /// Verifies a sample address against a canned successful Smarty response carrying the
+    /// given JSON body, returning the verification outcome.
+    /// </summary>
+    Task<Result<AddressUpdateSuccess>> ValidateAgainstCannedSuccessAsync(
```
noxmwalsh, line 18: This port is only here so the test-error endpoints can fake Smarty failures without hitting the network. 

Core is our business layer, so a Smarty canned-response API does would be probably not fit the best Can we move this out of `Core` and let `TestErrorController` talk to Infrastructure?

apps/portal/src/SEBT.Portal.Infrastructure/StateConnector/PluginEnrollmentCheckService.cs
```diff
@@ -0,0 +1,91 @@
...
+    private static PluginChildCheckRequest MapToPluginChild(ChildCheckRequest child) =>
+        new()
+        {
+            CheckId = child.CheckId,
+            FirstName = child.FirstName,
+            LastName = child.LastName,
+            DateOfBirth = child.DateOfBirth,
+            SchoolName = child.SchoolName,
+            SchoolCode = child.SchoolCode,
+            AdditionalFields = new Dictionary<string, string>(child.AdditionalFields)
```
noxmwalsh: Copying these dictionaries is fine and I get why it's being done here, but if child.AdditonalFields is `null`, `new Dictionary<>(null)`, I'm mostly sure it would throw here. Can we treat `null` as empty here and add a test to cover that case?

noxmwalsh (changes requested):
Changes requested; please see comments

noxmwalsh (approved):
The changes were addressed, so I'm approving this PR.  :shipit:

PR #590, opened 2026-08-17
DC-724(Chore): Next.js to 7; keep 6 as a side-by-side for API usage

noxmwalsh:
#### 🔗 Jira ticket
https://codeforamerica.atlassian.net/browse/DC-724

#### ✍️ Description

Supersedes Dependabot PR #574, which bumped TypeScript to 7.0.2 alone and failed CI with "Missing TypeScript package." Next.js 16.2.x checks for `typescript/lib/typescript.js`, which TypeScript 7 currently **does not** include, but will introduce a new API around October 2026 in 7.1. Next.js 16.3 resolves the project-local `tsc` binary instead and uses the Go-native compiler for everything else.

This PR upgrades Next.js and related packages to 16.3.1 across the portal, enrollment checker, design-system, and analytics workspaces. It adopts TypeScript 7.0.2 for compilation and CI typecheck while keeping the TypeScript 6 compiler API available for `typescript-eslint`, following [Microsoft's side-by-side guidance](https://devblogs.microsoft.com/typescript/announcing-typescript-7-0/#running-side-by-side-with-typescript-6-0):

#### 🔗 Links to related PRs

- Supersedes https://github.com/codeforamerica/sebt-self-service-portal/pull/574 (Dependabot TypeScript 7 bump)

#### ✅ Completion tasks

- [x] Added relevant tests (existing suites cover the change; no new behavior to unit test)
- [ ] Configuration changes:
  - [x] If new environment variables are added, update in Tofu (none)
  - [x] If new environment secrets are added, update in Tofu and set in AWS Secret Manager (none)
  - [x] If you're adding an appsetting, add it to the requisite example file, and update it in the AWS AppConfig (none)
  - [x] If appsetttings are changed, update in AWS AppConfig (none)

#### Test plan

For what I did to verify this locally:

- Portal and enrollment checker: `pnpm lint` (it should work against the TS 6 API)
- Portal and enrollment checker: `pnpm typecheck` (this should work against TS 7.0.2)
- Portal and enrollment checker: `pnpm build`
- Portal `Vitest` run.
- Enrollment checker `Vitest` run
- Design-system `Vitest` run
- Portal E2E (`pnpm ci:test:e2e`, DC, chromium)
- Backend unit tests (portal, state contract, CO connector, DC connector)

### DIFF
--- a/apps/portal/src/SEBT.EnrollmentChecker.Web/next.config.ts
+++ b/apps/portal/src/SEBT.EnrollmentChecker.Web/next.config.ts
@@ -24,6 +24,10 @@ const nextConfig: NextConfig = {
     NEXT_PUBLIC_BASE_PATH: basePath,
   },
   experimental: {
+    // @typescript/typescript6 (aliased as `typescript`) exposes tsc6, not tsc.
+    // Next's default CLI mode looks for typescript/bin/tsc; use the TS 6 API until
+    // typescript-eslint supports TS 7.1. CI `pnpm typecheck` still uses TS 7 via @typescript/native.
+    useTypeScriptCli: false,
     // Use our custom sass-loader configuration instead of built-in
     turbopackUseBuiltinSass: false
   },
--- a/apps/portal/src/SEBT.EnrollmentChecker.Web/package.json
+++ b/apps/portal/src/SEBT.EnrollmentChecker.Web/package.json
@@ -31,7 +31,7 @@
     "@tanstack/react-query": "^5.101.3",
     "@uswds/uswds": "^3.13.0",
     "i18next": "^26.3.4",
-    "next": "16.2.11",
+    "next": "16.3.1",
     "react": "19.2.7",
     "react-dom": "19.2.7",
     "react-i18next": "^17.0.7",
@@ -47,7 +47,7 @@
     "@types/react": "^19.2.17",
     "@types/react-dom": "^19.2.3",
     "eslint": "^9.39.4",
-    "eslint-config-next": "16.2.6",
+    "eslint-config-next": "16.3.1",
     "eslint-plugin-security": "^4.0.1",
     "jsdom": "^29.1.1",
     "msw": "^2.14.6",
@@ -57,7 +57,8 @@
     "sass": "^1.99.0",
     "sass-embedded": "^1.99.0",
     "sass-loader": "^17.0.0",
-    "typescript": "^6.0.3",
+    "@typescript/native": "npm:typescript@^7.0.2",
+    "typescript": "npm:@typescript/typescript6@^6.0.2",
     "vite": "^8.0.16",
     "vitest": "^4.1.6"
   }
--- a/apps/portal/src/SEBT.Portal.Web/next.config.ts
+++ b/apps/portal/src/SEBT.Portal.Web/next.config.ts
@@ -26,6 +26,10 @@ const nextConfig: NextConfig = {
     NEXT_PUBLIC_SMARTY_EMBEDDED_KEY: process.env.NEXT_PUBLIC_SMARTY_EMBEDDED_KEY || ''
   },
   experimental: {
+    // @typescript/typescript6 (aliased as `typescript`) exposes tsc6, not tsc.
+    // Next's default CLI mode looks for typescript/bin/tsc; use the TS 6 API until
+    // typescript-eslint supports TS 7.1. CI `pnpm typecheck` still uses TS 7 via @typescript/native.
+    useTypeScriptCli: false,
     // Use our custom sass-loader configuration instead of built-in
     turbopackUseBuiltinSass: false,
     // Tree-shake the design-system barrel so importing a single export
--- a/apps/portal/src/SEBT.Portal.Web/package.json
+++ b/apps/portal/src/SEBT.Portal.Web/package.json
@@ -34,7 +34,7 @@
   },
   "dependencies": {
     "@amplitude/analytics-browser": "^2.42.3",
-    "@next/third-parties": "^16.2.6",
+    "@next/third-parties": "^16.3.1",
     "@opentelemetry/api": "^1.9.1",
     "@sebt/analytics": "workspace:*",
     "@sebt/design-system": "workspace:*",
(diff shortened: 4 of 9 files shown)


adbergen (changes requested):
Checked the aliasing mechanics, useTypeScriptCli is a real Next 16.3 option that defaults to true, the tsc6/tsc bin split is unambiguous, and typescript-eslint binds to the TS 6 shim in the lockfile. Approach is sound. Also worth noting ^6.0.2 isn't the downgrade from ^6.0.3 it looks like. The shim depends on "@typescript/old": "npm:typescript@^6", so the real compiler still floats to 6.0.3.

Blocker — rebase. Conflicts across all five package.json files plus the lockfile, 33 commits behind, CI last ran Aug 17. Main independently landed Next 16.3.0 in the meantime, so the unique Next delta is now just a patch. Stay on 16.3.1 — 16.3.2 and 16.3.3 are both still inside the 7-day cooldown. Also needs the all states label.

While you're in there, the 16 cooldown exemptions are all unnecessary now. The two TypeScript entries were never needed (both published early July, six weeks before this opened); the Next 16.3.1 family aged out around Aug 20. Inert either way, so not a security hold, but free to drop during a rebase you're doing anyway, and check:cooldown-exclusions will never flag them, since it only detects "no longer resolves," not "no longer needed."

Maybe follow-up ticket? The next.config.ts comment already names the trigger (typescript-eslint on TS 7.1, ~Oct). Worth a real ticket to unwind the shim, the @typescript/native alias, and useTypeScriptCli: false — otherwise an experimental flag and an inverted alias become permanent by default. Two things to fold into it: @typescript/native isn't a real package name (Microsoft's is @typescript/native-preview), and next build now typechecks with TS 6 while CI typechecks with TS 7 — fine, but nothing signals it.

One question: typescript was added to packages/analytics and packages/design-system, which never had it and have no typecheck script. Deliberate, to keep one TS across the workspace so a transitive peer can't pull a second copy? If so it needs a comment — it currently reads as an unused devDependency.

noxmwalsh:
> Checked the aliasing mechanics, useTypeScriptCli is a real Next 16.3 option that defaults to true, the tsc6/tsc bin split is unambiguous, and typescript-eslint binds to the TS 6 shim in the lockfile. Approach is sound. Also worth noting ^6.0.2 isn't the downgrade from ^6.0.3 it looks like. The shim depends on "@typescript/old": "npm:typescript@^6", so the real compiler still floats to 6.0.3.
> 
> Blocker — rebase. Conflicts across all five package.json files plus the lockfile, 33 commits behind, CI last ran Aug 17. Main independently landed Next 16.3.0 in the meantime, so the unique Next delta is now just a patch. Stay on 16.3.1 — 16.3.2 and 16.3.3 are both still inside the 7-day cooldown. Also needs the all states label.
> 
> While you're in there, the 16 cooldown exemptions are all unnecessary now. The two TypeScript entries were never needed (both published early July, six weeks before this opened); the Next 16.3.1 family aged out around Aug 20. Inert either way, so not a security hold, but free to drop during a rebase you're doing anyway, and check:cooldown-exclusions will never flag them, since it only detects "no longer resolves," not "no longer needed."
> 
> Maybe follow-up ticket? The next.config.ts comment already names the trigger (typescript-eslint on TS 7.1, ~Oct). Worth a real ticket to unwind the shim, the @typescript/native alias, and useTypeScriptCli: false — otherwise an experimental flag and an inverted alias become permanent by default. Two things to fold into it: @typescript/native isn't a real package name (Microsoft's is @typescript/native-preview), and next build now typechecks with TS 6 while CI typechecks with TS 7 — fine, but nothing signals it.
> 
> One question: typescript was added to packages/analytics and packages/design-system, which never had it and have no typecheck script. Deliberate, to keep one TS across the workspace so a transitive peer can't pull a second copy? If so it needs a comment — it currently reads as an unused devDependency.

I don't think it's required for those two. From what I see, `Observability` already has typescript declared in `package.json`; I just added the TS 6 shim to analytics and design-system so every workspace package resolved the same alias. 

I’ll remove it from those two and keep the pin on the apps and observability.

noxmwalsh:
> > Checked the aliasing mechanics, useTypeScriptCli is a real Next 16.3 option that defaults to true, the tsc6/tsc bin split is unambiguous, and typescript-eslint binds to the TS 6 shim in the lockfile. Approach is sound. Also worth noting ^6.0.2 isn't the downgrade from ^6.0.3 it looks like. The shim depends on "@typescript/old": "npm:typescript@^6", so the real compiler still floats to 6.0.3.
> > Blocker — rebase. Conflicts across all five package.json files plus the lockfile, 33 commits behind, CI last ran Aug 17. Main independently landed Next 16.3.0 in the meantime, so the unique Next delta is now just a patch. Stay on 16.3.1 — 16.3.2 and 16.3.3 are both still inside the 7-day cooldown. Also needs the all states label.
> > While you're in there, the 16 cooldown exemptions are all unnecessary now. The two TypeScript entries were never needed (both published early July, six weeks before this opened); the Next 16.3.1 family aged out around Aug 20. Inert either way, so not a security hold, but free to drop during a rebase you're doing anyway, and check:cooldown-exclusions will never flag them, since it only detects "no longer resolves," not "no longer needed."
> > Maybe follow-up ticket? The next.config.ts comment already names the trigger (typescript-eslint on TS 7.1, ~Oct). Worth a real ticket to unwind the shim, the @typescript/native alias, and useTypeScriptCli: false — otherwise an experimental flag and an inverted alias become permanent by default. Two things to fold into it: @typescript/native isn't a real package name (Microsoft's is @typescript/native-preview), and next build now typechecks with TS 6 while CI typechecks with TS 7 — fine, but nothing signals it.
> > One question: typescript was added to packages/analytics and packages/design-system, which never had it and have no typecheck script. Deliberate, to keep one TS across the workspace so a transitive peer can't pull a second copy? If so it needs a comment — it currently reads as an unused devDependency.
> 
> I don't think it's required for those two. From what I see, `Observability` already has typescript declared in `package.json`; I just added the TS 6 shim to analytics and design-system so every workspace package resolved the same alias.
> 
> I’ll remove it from those two and keep the pin on the apps and observability.

Addressed in 54fbe2178560b8b05eed0ecfaca369adf11836db

adbergen (approved):

PR #608, opened 2026-08-20
[DC-706] Add unit coverage for OIDC code exchange

adbergen:
#### 🔗 Jira ticket
https://codeforamerica.atlassian.net/browse/DC-706

#### ✍️ Description
`OidcExchangeService`'s main paths were at 0% unit coverage — the existing tests only cover the discovery HTTPS check and `GetDiscoveryConfigAsync`. This adds 15 unit tests for `ExchangeCodeAsync` and (indirectly) the private `EnrichClaimsFromUserInfo`, with **no AspNetCore host and no live IdP**:

- **Discovery + JWKS** are served by a local HTTP stand-in signed with an RSA key the test owns, so the id_token verification (JWKS + issuer + audience + lifetime) runs for real. Each test uses a unique realm so the service's process-wide discovery-config cache never collides.
- **Token + userinfo** HTTP calls are mocked through `IHttpClientFactory`.
- `EnrichClaimsFromUserInfo` is private, so it's exercised through `ExchangeCodeAsync`.

**Coverage — maps to the acceptance criteria:**
- **Happy path:** valid code → signed callback token carrying id_token identity claims (sub/email) + userinfo-enriched claims (givenName/familyName/name/phone).
- **Failure cases:** not-configured (503), discovery load failure (502), discovery missing `token_endpoint` (502), token request throws, IdP rejects the code (400 `invalid_grant`), non-JSON token body, no `id_token`, expired id_token, wrong issuer, and no resolvable sub/email.
- **Enrichment success/failure:** recovers email from the userinfo `email` field, recovers email from `preferred_username` when it looks like an email, and both failure modes (non-success status + malformed JSON) are non-fatal.

End-to-end coverage of the controller callback (and the DC-702 host refactor) is intentionally left to E2E tests, per the ticket.

#### 🔗 Links to related PRs
None. (Related refactor: DC-702, tracked separately.)

#### ✅ Completion tasks
- [x] Added relevant tests — 15 new unit tests; 21/21 `OidcExchangeService` unit tests (existing 6 + new 15) pass, build clean.
- [x] Meets acceptance criteria in ticket — `ExchangeCodeAsync` happy + key failure paths and `EnrichClaimsFromUserInfo` success/failure are covered; the exchange/enrichment methods are no longer at 0%. No host or live IdP required (HTTP mocked).
- [ ] Configuration changes: **N/A** — test-only change; no environment variables, secrets, or appsettings.

PR #617, opened 2026-08-25
[DC-702] Move OIDC orchestration from API into UseCases

adbergen:
#### 🔗 Jira ticket
https://codeforamerica.atlassian.net/browse/DC-702

#### ✍️ Description
OIDC was the one product flow that didn't dispatch UseCase handlers — `OidcController` orchestrated the whole callback/complete-login flow inline (including injecting `IUserRepository` for user upsert), and the exchange service lived in the Api layer. This moves the business orchestration into UseCases behind Core ports, per Clean Architecture (docs/adr/0002).

**How each acceptance criterion is met:**

- **Orchestration in UseCases** — two new handlers:
  - `OidcCallbackCommandHandler`: session enforcement, state (CSRF) validation, replay protection, code exchange, session advancement.
  - `CompleteOidcLoginCommandHandler`: session advancement/consumption, callback-token verification, user resolution (existing-only for step-up, get-or-create for login), portal JWT minting.
- **Controllers dispatch handlers only** — `OidcController` no longer injects `IUserRepository` or `IOidcTokenService`; `Callback`/`CompleteLogin` build a command from the request + `oidc_session` cookie, dispatch via `ICommandHandler<…>` (same pattern as `OtpController`), and map `Result<T>` subtypes to the exact HTTP contract as before (`MapFailure`: Forbidden→403, NotConfigured→503, ConnectionFailed→502, everything else→400). `AuthController` was already compliant.
- **PKCE, cookies, origin middleware, HTTP mapping stay in API** — `GetConfig`/`Authorize`/`report-failure` are substantively untouched; `OidcSessionCookie`/`AuthCookies` reads/writes and status mapping are now the controller's entire job. `OidcOriginValidationMiddleware` untouched.
- **UseCase unit tests without the AspNetCore host** — 12 callback + 16 complete-login handler tests (NSubstitute mocks only) covering every success and failure path; controller tests re-pointed at mocked handlers; `ValidateCallbackToken` gets real-crypto round-trip/expired/garbage/not-configured tests.
- **Behavior unchanged** — same routes, status codes, response bodies, failure-log lines, and session one-shot semantics. The integration behavior locks (`OidcPreAuthSecurityTests`, `LogoutTokenDenylistTests`) pass with only using/mock-type updates.

**Structural moves:**
- Ports → Core: `IOidcExchangeService` (with a web-free `OidcExchangeResult` — `FailureReason` enum instead of an HTTP `StatusCode` — and a narrow `OidcDiscoveryInfo` so Core doesn't take the OIDC protocol library dependency), `IPreAuthSessionStore` + `PreAuthSession`, `IOidcCallbackFailureLogger` + log entry, `OidcLogSanitizer`.
- Adapter → Infrastructure: `OidcExchangeService` implementation (mirrors `JwtTokenService`), gaining `ValidateCallbackToken()` so callback-token signing and validation share one key/issuer/audience implementation.
- Deleted `JwtTokenService`'s parallel copy of the OIDC infrastructure claim set — both consumers now use `Core.OidcClaims.InfrastructureClaimNames` (the duplication existed only because of the old layering).
- Store/failure-logger implementations stay in Api (session/cookie plumbing); DI unchanged apart from the exchange service now resolving from Infrastructure.
- Test relocations per the namespace-mirroring convention: exchange tests → `Unit/Infrastructure/Services/`, sanitizer tests → `Unit/Core/Utilities/`.

**Two deliberate deviations (flagging for review):**
1. Two step-up 400 responses (`step_up_user_not_found`, `token_generation_failed`) previously returned anonymous `{ "error": … }` bodies; they now serialize through the shared `ErrorResponse` record, so they gain an inert `"errors": null` field. Every other failure body on these endpoints already used `ErrorResponse`.
2. The success/failure log lines that moved into handlers keep identical message content and structured fields, but their logger *category* changes from `OidcController` to the handler types. The `OidcCallbackFailureLogger` off-boarding lines (the ones Datadog dashboards key on) are unchanged, category included.

Net −751 lines in src.

#### 🔗 Links to related PRs
None. (Test groundwork landed in #608 / DC-706.)

#### ✅ Completion tasks
- [x] Added relevant tests — 28 new handler unit tests + controller mapping tests + 4 `ValidateCallbackToken` crypto tests; full suite: 1,975 tests, 0 failures (integration OIDC security suites included).
- [x] Meets acceptance criteria in ticket — each AC mapped above.
- [ ] Configuration changes: **N/A** — no environment variables, secrets, or appsettings changes; DI wiring only (same interfaces, same lifetimes).


apps/portal/src/SEBT.Portal.UseCases/Auth/OidcCallback/OidcCallbackCommandHandler.cs
```diff
@@ -0,0 +1,146 @@
...
+public class OidcCallbackCommandHandler(
+    IPreAuthSessionStore sessionStore,
+    IOidcExchangeService exchangeService,
+    IOidcCallbackFailureLogger callbackFailureLogger,
+    ILogger<OidcCallbackCommandHandler> logger)
+    : ICommandHandler<OidcCallbackCommand, OidcCallbackResponse>
+{
+    // HTTP statuses recorded on off-boarding log entries; pass-through diagnostic data
+    // matching what the API layer returns for the corresponding result type.
+    private const int StatusForbidden = 403;
```
noxmwalsh: The handlers log `HttpStatus` for Datadog before the API actually returns these statuses. If we later change the API mapping, the dashboard would show the old code and while datadog would count the wrong failures (or vice versa) because it's in two place currently.

We should either log the actual reason (`missing_session`, `replay`etc.) here, or set `HttpStatus` only from the status we actually returned if we can.  

I'm not going to block on this since the mapping still match correctly, but the OIDC logs are critical enough where I think preventing a future drift issue is worth looking into.

apps/portal/src/SEBT.Portal.Infrastructure/SEBT.Portal.Infrastructure.csproj
```diff
@@ -43,6 +43,7 @@
       <PackageReference Include="Microsoft.Extensions.Options" Version="10.0.10" />
       <PackageReference Include="Microsoft.Extensions.Options.ConfigurationExtensions" Version="10.0.10" />
       <PackageReference Include="Microsoft.FeatureManagement" Version="4.5.0" />
+      <PackageReference Include="Microsoft.IdentityModel.Protocols.OpenIdConnect" Version="8.19.2" />
```
noxmwalsh, line 46: If we move this package into Infrastructure now (which is good!), should we remove the OpenIdConnect package from API's `csproj`?

noxmwalsh (approved):
Non-blocking comments added, but I think this is fine to merge in

PR #618, opened 2026-08-25
[DC-499] Answer anonymous auth-status probes with 200 instead of 401

adbergen:
#### 🔗 Jira ticket
https://codeforamerica.atlassian.net/browse/DC-499

#### ✍️ Description
The SPA probes `GET /api/auth/status` on every page load — including logged-out ones — to decide whether to show the login UI. The endpoint was `[Authorize]`, so every anonymous visitor (or bot) produced a 401: ~2k/day in CO prod frontend logs, plus an OTel ERROR span in the Next proxy and a red console line in the browser. "Not signed in" is an expected answer to this probe, not an error.

**The contract already anticipated the fix**: the client's `fetchSession` has always handled `isAuthorized: false` (previously unreachable — the response doc literally called the flag "always-true"), and every other response field was already nullable. This change makes the backend able to say `false`:

- `/api/auth/status` is now `[AllowAnonymous]`; an unauthenticated principal gets **200 `{ isAuthorized: false }`** with no claims. The authenticated path is byte-identical. The anonymous branch intentionally logs nothing.
- **Client logic is unchanged** — the `false` path already existed. The 401 catch is kept (and kept under test) so older API deployments still read as "not logged in" during a rollout window.

**Security invariants preserved, and proven by tests:**
- Bearer authentication still runs before the handler, so an expired/tampered/policy-violating cookie is **still cleared via Set-Cookie** (`GetStatus_WithSessionCookieMissingAuthTime_ClearsCookie` now asserts 200 + `isAuthorized: false` + Set-Cookie).
- Tokens revoked at logout yield the anonymous shape with **no claims**, via both the cookie and the Authorization header (denylist tests strengthened from status-code asserts to body asserts).
- Every other endpoint remains `[Authorize]`; no new information is exposed — a 401 already told anonymous callers exactly this.

Effect: anonymous page views are 200s end-to-end, eliminating the 401 noise at the source (frontend logs, proxy spans, browser console).

#### 🔗 Links to related PRs
None. (Will trivially merge around #617, which touches a different region of AuthController.)

#### ✅ Completion tasks
- [x] Added relevant tests — new anonymous-branch unit test; integration tests flipped to the new contract and strengthened to assert the no-claims body; legacy-401 client handling kept under test. Full suites green: 1,949 backend unit + 58 auth integration + 1,075 frontend.
- [x] Meets acceptance criteria in ticket — anonymous portal visits no longer surface 401 errors client-side or in logs.
- [ ] Configuration changes: **N/A** — no environment variables, secrets, or appsettings.


apps/portal/src/SEBT.Portal.Api/Models/AuthorizationStatusResponse.cs
spokenbird: Minor nit: After this PR the email can be null so worth updating this doc to mention as much.

spokenbird (approved):

PR #640, opened 2026-08-31
Guard push-button releases from overwriting a published same-day release

spokenbird:
#### ✍️ Description
`push-button-release.yaml`'s release tag is date + state based
(`YYYY.MM.DD-dc`/`YYYY.MM.DD-colorado`), and `action-gh-release` upserts by
tag. A second same-day run for a state would silently rewrite an
already-published release's notes in place, not just an unpublished draft.

This adds a tag-resolution step before creating the release:
- If a release for today's tag already exists and is still a **draft**,
  reuse it (unchanged behavior — lets you iterate before publishing).
- If it's already **published**, don't touch it — walk to the next available
  suffix (`-2`, `-3`, ...) instead, so multiple real releases can still be
  cut for the same state on the same day.
- The release **name** gets a matching `(N)` suffix in the escalated case, so
  two same-day releases don't render with an identical title.


necampanini (approved):
looks good!

PR #659, opened 2026-09-02
Chore: serve Next.js client config at runtime instead of inlining at build

adbergen:
#### 🔗 Jira ticket

"Make Next.js client config available at runtime"

#### ✍️ Description

The portal inlined browser-facing config into the client bundle at `pnpm build`, so setting those variables on a running container or in IIS `web.config` could not change what the browser received. That made one build artifact un-promotable across environments, breaking twelve-factor [Config (III)](https://12factor.net/config) and [Build, release, run (V)](https://12factor.net/build-release-run).

Browser config is now resolved **at request time**. Each app uses the mechanism that fits how it is served:

**Portal — server-injected.** Nine variables dropped the `NEXT_PUBLIC_` prefix so Next stops inlining them. The root layout resolves them per request (`lib/runtime-config.ts`) and hands them to the client through `RuntimeConfigProvider`; client components read `useRuntimeConfig()`. This costs no prerendering — `generateMetadata` already awaits `headers()`, so only `_global-error` and `favicon.ico` were ever prerendered.

**Enrollment checker — runtime `config.js`.** The checker ships as a static export (`output: 'export'`) to S3, so there is no server to read env at request time. The deployed bucket carries a `config.js` assigning `window.__CHECKER_CONFIG__`, loaded `beforeInteractive` so values are available on first paint. Fourteen variables moved. The analytics and pixel tags became a client component — rendered from a Server Component they would be frozen into the exported HTML, which is the bug itself.

**CSP moves with it.** `proxy.ts` gates `connect-src`/`script-src` on whether an analytics key is present, and those reads were inlined too. Previously a key supplied at runtime would still have been blocked by a policy frozen at build time. The middleware now reads the same unprefixed names per request.

**The state moves to runtime too, which required changing how theming is built.** `STATE` was not merely a value — it selected the compiled USWDS stylesheet, the `next/font` faces, and the locale bundle. Sass configures `uswds-core` once per compilation (`@use ... with (...)` is singleton), so two themes cannot coexist in one stylesheet. Each state is now compiled to its own file (`public/themes/theme-{state}.css`, ~630 KB) and linked once `STATE` is known at request time, so a visitor downloads one theme rather than every state's. Every state's fonts ship in one generated module with `preload` off. `getState()` now resolves `<html data-state>` first — the browser's only runtime source, stamped by the server per request — which kept all 27 of its call sites unchanged.

**The portal image is state-agnostic.** The Dockerfile no longer takes `ARG STATE` and neither ECR build passes it; the ECS task definition supplies it. Each environment still pushes to its own ECR repository — consolidating those is infrastructure work outside this repo.

Only build-*identity* values remain inlined (`NEXT_PUBLIC_BUILD_SHA`, `NEXT_PUBLIC_DC_CONNECTOR_SHA`) — they describe the artifact, not the environment. Rationale and rejected alternatives are in [ADR 0022](docs/adr/0022-runtime-client-config.md).

**Verified end to end, not just asserted:**

- Portal config: one artifact built with **no** client config served `G-RUNTIME-AAA`/`smarty-aaa` on one run and `G-RUNTIME-BBB`/`smarty-bbb`/`amp-bbb` on the next, with the CSP widening for Amplitude only on the second.
- Portal state: the same artifact, built with **no `STATE` at all**, served `data-state="dc"` + `theme-dc.css` + *"District of Columbia SUN Bucks Self-Service Portal"*, then `data-state="co"` + `theme-co.css` + *"Colorado Summer EBT Self-Service Portal"* — purely from `STATE`, no rebuild. The two stylesheets are genuinely distinct (DC buttons `#2a646d`, CO `#168092`).
- Checker: static export built with **zero** client-config env; swapping only `config.js` changed the served config with `index.html` unchanged (no rebuild).
- Portal 1086 unit tests and 67 E2E pass, checker 246 pass, design-system 203 pass; typecheck and lint clean throughout. The Amharic screenshot test still passes, confirming the CSS-delivery change did not shift rendering.

#### ⚠️ Environment variable changes — action required before merge

**This cannot ship until the values exist where the process runs.** The build args are gone, so merging without the steps below turns analytics, Socure, and Smarty **off** in deployed environments.

**1. Renamed — drop the `NEXT_PUBLIC_` prefix (portal, both deploy paths).** Same values, new names:

| Was (build arg) | Now (runtime env) |
| --- | --- |
| `NEXT_PUBLIC_GA_ID` | `GA_ID` |
| `NEXT_PUBLIC_AMPLITUDE_API_KEY` | `AMPLITUDE_API_KEY` |
| `NEXT_PUBLIC_MIXPANEL_TOKEN` | `MIXPANEL_TOKEN` |
| `NEXT_PUBLIC_SITEIMPROVE_ID` | `SITEIMPROVE_ID` |
| `NEXT_PUBLIC_SOCURE_DI_SDK_KEY` | `SOCURE_DI_SDK_KEY` |
| `NEXT_PUBLIC_SMARTY_EMBEDDED_KEY` | `SMARTY_EMBEDDED_KEY` |
| `NEXT_PUBLIC_MOCK_SOCURE` | `MOCK_SOCURE` |
| `NEXT_PUBLIC_DEBUG_REPEAT_OIDC_STEP_UP` | `DEBUG_REPEAT_OIDC_STEP_UP` |
| `NEXT_PUBLIC_SOCURE_SDK_KEY` | `SOCURE_SDK_KEY` (declared; currently unread) |
| `NEXT_PUBLIC_STATE` | `STATE` |

**2. Docker / ECR path (DC dev, CO) — Tofu.** These must be set on the **ECS task definition** for the web container. They are no longer passed as `--build-arg` in `deploy-ecr.yaml`. **`STATE` is now required there** — without it the container defaults to `dc`, so the CO task definition must set `STATE=co` explicitly.

**3. IIS path (DC prod) — `web.config`.** `scripts/ci/templates/web.config` now carries an `<environmentVariables>` block with empty placeholders for each. Fill these in on the target host per environment. This block previously had no effect on browser code; it does now. `NEXT_PUBLIC_STATE` there was renamed to `STATE`.

**4. Enrollment checker — no infra change.** Both deploy jobs write `config.js` into the export from the existing GitHub vars (`ENROLLMENT_CHECKER_API_BASE_URL`, `ENROLLMENT_CHECKER_PORTAL_URL`, `ENROLLMENT_CHECKER_APPLICATION_URL`, `AMPLITUDE_API_KEY`, `MIXPANEL_TOKEN`, `SITEIMPROVE_ID`). No new GitHub variables are introduced anywhere.

**5. No new secrets.** Every value here is browser-facing by definition and already reaches the client. Nothing moves into or out of AWS Secret Manager.

#### 🔗 Links to related PRs

None.

#### ✅ Completion tasks

- [x] Added relevant tests
- [ ] Configuration changes:
  - [ ] If new environment variables are added, update in Tofu — **required: the eight renamed vars must be set on the ECS task definition (see above)**
  - [x] If new environment secrets are added, update in Tofu and set in AWS Secret Manager — none; all values are browser-facing
  - [x] If you're adding an appsetting, add it to the requisite example file, and update it in the AWS AppConfig — none
  - [x] If appsetttings are changed, update in AWS AppConfig — none


apps/portal/src/SEBT.EnrollmentChecker.Web/src/lib/applyHref.ts
```diff
@@ -23,7 +25,7 @@ const PEAK_LANG_BY_LOCALE: Record<string, string> = {
 }
 
 export function getApplyHref(locale: string): string | null {
-  const base = process.env.NEXT_PUBLIC_APPLICATION_URL
+  const base = getClientConfig().applicationUrl
```
spokenbird, line 30: I think there is a regression here, which is that outside of local development, the env.ts zod validation never runs against this since it's now sourced from config.js and the env fallback is only used when config.js doesn't set it.  This would mean rather than getting a build-time error calling out the invalid URL, we would get a runtime error.

apps/portal/src/SEBT.Portal.Web/src/features/auth/components/IalGuard/IalGuard.tsx
spokenbird: Stale comment still referenceds NEXT_PUBLIC, we may wanna do a quick pass to find any other stale sites.

CLAUDE.md
spokenbird: This is now incorrect I believe with the changes to next.config.ts and the sassOptions block there?

apps/portal/src/SEBT.EnrollmentChecker.Web/src/lib/client-config.ts
```diff
@@ -0,0 +1,87 @@
...
+    showSchoolField: bool(o.showSchoolField, env.NEXT_PUBLIC_SHOW_SCHOOL_FIELD),
+    checkerEnabled: bool(o.checkerEnabled, env.NEXT_PUBLIC_CHECKER_ENABLED),
+    botProtectionEnabled: bool(o.botProtectionEnabled, env.NEXT_PUBLIC_BOT_PROTECTION_ENABLED),
+    amplitudeApiKey: str(o.amplitudeApiKey, env.NEXT_PUBLIC_AMPLITUDE_API_KEY),
+    mixpanelToken: str(o.mixpanelToken, env.NEXT_PUBLIC_MIXPANEL_TOKEN),
+    siteImproveId: str(o.siteImproveId, env.NEXT_PUBLIC_SITEIMPROVE_ID),
+    metaPixel: str(o.metaPixel, env.NEXT_PUBLIC_META_PIXEL),
+    metaPixelAction: str(o.metaPixelAction, env.NEXT_PUBLIC_META_PIXEL_ACTION),
+    adentifiPixelLanding: str(o.adentifiPixelLanding, env.NEXT_PUBLIC_ADENTIFI_PIXEL_LANDING),
+    adentifiPixelApplyNow: str(o.adentifiPixelApplyNow, env.NEXT_PUBLIC_ADENTIFI_PIXEL_APPLY_NOW)
```
spokenbird, line 107: I think we may need to thread some of these through CI? In particular showSchoolField, checkerEnabled, botProtectionEnabled, gaId, metaPixel, metaPixelAction, adentifiPixelLanding, adentifiPixelApplyNow are not in our github env vars right now. They'd also need to be added to the deploy-enrollment-checker workflow in the `Write runtime client config` step.

Not sure which we want, vs which can be omitted (I imagine we don't want the meta pixel in our development environments for example).

apps/portal/src/SEBT.Portal.Web/src/features/auth/components/doc-verify/adapters/index.ts
```diff
@@ -6,8 +6,8 @@ import { MockDocVAdapter } from './mock-adapter'
 import { SocureDocVAdapter } from './socure-adapter'
 import type { DocVAdapter } from './types'
 
-export function createDocVAdapter(): DocVAdapter {
-  if (process.env.NEXT_PUBLIC_MOCK_SOCURE === 'true') {
+export function createDocVAdapter(mockSocure: boolean): DocVAdapter {
```
spokenbird: I can't find any call sites for this, is it dead code?

.github/workflows/deploy-enrollment-checker.yaml
```diff
@@ -56,13 +56,24 @@ jobs:
...
           NEXT_PUBLIC_BUILD_SHA: ${{ github.sha }}
-          NEXT_PUBLIC_AMPLITUDE_API_KEY: ${{ vars.AMPLITUDE_API_KEY }}
-          NEXT_PUBLIC_MIXPANEL_TOKEN: ${{ vars.MIXPANEL_TOKEN }}
-          NEXT_PUBLIC_SITEIMPROVE_ID: ${{ vars.SITEIMPROVE_ID }}
+
+      - name: Write runtime client config
+        # Browser config lives in the deployed bucket, not the bundle, so the same
+        # export can be promoted between environments by rewriting this one file.
+        run: |
+          cat > apps/portal/src/SEBT.EnrollmentChecker.Web/out/config.js <<'EOF'
```
spokenbird: I'd recommend using an env block here rather than interpolating directly, paired with a JSON stringify, something like:


```
- name: Write runtime client config
  env:
    API_BASE_URL: ${{ vars.ENROLLMENT_CHECKER_API_BASE_URL }}
    PORTAL_URL: ${{ vars.ENROLLMENT_CHECKER_PORTAL_URL }}
    APPLICATION_URL: ${{ vars.ENROLLMENT_CHECKER_APPLICATION_URL }}
    AMPLITUDE_API_KEY: ${{ vars.AMPLITUDE_API_KEY }}
    MIXPANEL_TOKEN: ${{ vars.MIXPANEL_TOKEN }}
    SITEIMPROVE_ID: ${{ vars.SITEIMPROVE_ID }}
  run: |
    node -e "
      const config = {
        apiBaseUrl: process.env.API_BASE_URL,
        portalUrl: process.env.PORTAL_URL,
        applicationUrl: process.env.APPLICATION_URL,
        amplitudeApiKey: process.env.AMPLITUDE_API_KEY,
        mixpanelToken: process.env.MIXPANEL_TOKEN,
        siteImproveId: process.env.SITEIMPROVE_ID
      };
      require('fs').writeFileSync(
        'apps/portal/src/SEBT.EnrollmentChecker.Web/out/config.js',
        'window.__CHECKER_CONFIG__ = ' + JSON.stringify(config, null, 2) + ';\n'
      );
    "
```

If you wanted to go the extra mile we could also maybe extract to a seperate script thats called for both DC and CO?

spokenbird (changes requested):
A few smaller things called out directly in comments on the files themselves, but two big things I noticed:

1. `apps/portal/src/SEBT.Portal.Web/src/lib/i18n-init.ts` wasn't migrated to the new STATE pattern, client-side i18n will silently fall back to DC in CO.

`const state = (process.env.NEXT_PUBLIC_STATE || process.env.STATE || 'dc').toLowerCase()` will fail since we no longer pass NEXT_PUBLIC_STATE and STATE isn't in the client, it falls back to DC. The fix would be updating this site to use the net getState() helper. 

2. We need to add the tofu changes for all the environment variables that were migrated from the NEXT_PUBLIC pattern. I believe that's GA_ID, AMPLITUDE_API_KEY, MIXPANEL_TOKEN, SITEIMPROVE_ID, SOCURE_SDK_KEY, SOCURE_DI_SDK_KEY, SMARTY_EMBEDDED_KEY, and MOCK_SOCURE, but of this set, I'm not 100 percent on which actually need to be set in our AWS development environments? I can help with actually setting them if desired, I just need to know which ones to set for which environments.

spokenbird (approved):

PR #662, opened 2026-09-02
Sync CO content from working sheet, preserve closed-season copy

spokenbird:
#### ✍️ Description
Syncs `co.csv` with the latest from the content working sheet and reconciles the drift introduced from hardcoding changes that were not in the content working sheet:

- 5 keys had identical copy moved under new names in the sheet (e.g.
  `apply2027Action` → `applyForSebtClosedBody3`) — code updated to match.
- 2 keys were dropped from the sheet entirely but are still read by code
  (`apply2027StepNote`, `GLOBAL - Processing`) — re-added as-is.
- 30 keys where the new sheet reverted DC-701 closed-season copy back to
  open-season wording — reverted to keep the closed-season copy live,
  confirmed that closed season wording in all places is correct.

No component logic changed beyond the 7 translation-key renames.


apps/portal/src/SEBT.Portal.Web/src/features/household/components/BenefitExpirationBanner/BenefitExpirationBanner.tsx
```diff
@@ -5,7 +5,7 @@ import { useTranslation } from 'react-i18next'
 import { useFeatureFlag } from '@/features/feature-flags'
 import { Alert } from '@sebt/design-system'
 
-// Keys map to CSV: "S2 - Portal Dashboard - Alert Benefit Expiration - {Key}"
+// Keys map to CSV: "S2 - Portal Dashboard - Alert {Title,Body} Expungement"
```
noxmwalsh, line 8: This component is shared with DC, so any changes reflect that as well

noxmwalsh (approved):
I think this is good, but I want to point out there's a change that affected DC as well

apps/portal/src/SEBT.Portal.Web/src/features/household/components/BenefitExpirationBanner/BenefitExpirationBanner.tsx
spokenbird: Confirmed the Title key text is identical for both versions in DC and the Body key no longer exists as the old version in the DC content manager sheet. It only exists under `S2 - Portal Dashboard - Alert Body Expungement` and the only difference is the word child changes to student in `Check each enrolled student here to see when their benefits expire.`

PR #674, opened 2026-09-08
DC-742(Feature): Hide card replacement when benefits are near expiration

noxmwalsh:
#### 🔗 Jira ticket
[DC-742](https://codeforamerica.atlassian.net/browse/DC-742)

#### ✍️ Description
Adds an optional per-state cutoff that turns off order-new-card for a case once remaining calendar days until `BenefitExpirationDate` are less than or equal to `SelfServiceRules:CardReplacement:DisableDaysBeforeExpiration`.

The cutoff is independent of the statewide `CardReplacement.Enabled` kill switch. Statewide off still denies every card. When statewide is on, the cutoff still applies per case. Address update is unchanged. A missing expiration date does not trigger the cutoff.

In the case of mixed households, the expectation is that we'll keep the household "request-new-cards" CTA if any child is still eligible. Per-case `allowCardReplacement` hides replacement links, picker rows, and the `/cards/replace` deep link (which redirects to the dashboard).

The new setting is omitted from `appsettings.json` so current behavior stays the default (mainly so DC doesn't have to have setting present either way). Commented examples for the new setting lives in `appsettings.dc.example.json` and `appsettings.co.example.json`.

#### 🔗 Links to related PRs
None

#### ✅ Completion tasks

- [x] Added relevant tests
- [ ] Configuration changes:
  - [x] If you're adding an appsetting, add it to the requisite example file, and update it in the AWS AppConfig

#### Test plan

Local mock data only. Uncomment `DisableDaysBeforeExpiration` in `appsettings.dc.json` or `appsettings.co.json` (not the example file), set it to `4`, keep `UseMockHouseholdData` true, and restart the API. Remaining calendar days of 4 or fewer are denied; 5 or more stay allowed. The mixed mock households use remaining 3, 4, and 10 days, so two cards should hide replacement and one should keep it.

DC:
- Sign in as `sebt.dc+summer-active@codeforamerica.org` (OTP from Mailpit).
- Dashboard: household Request-new-cards CTA is visible. Noah (3 days) and Mia (4 days) have no per-card replacement link. Luna (10 days) still does.
- Open replacement for Luna and confirm the flow still works. Address update still works for the household.
- Deep-link `/cards/replace?case=` for Noah or Mia; the app should send you to the dashboard, not the confirm-address step.

CO (Make sure Keycloak is running):
- Route lookup to the CoActive persona: set `DevelopmentPhoneOverride:Phone` to `3035551009`, or sign in with a user whose phone claim is that number. If a leftover DC session skips the OIDC exchange, log out first (`/api/auth/logout`) and confirm in Keycloak.
- Same mixed-household checks as DC: Camila (3 days, last four 1234) and Mateo (4 days, 2345) off; Sofia (10 days, 3456) on. Household CTA stays. Address update should still work; denied deep links should go to the dashboard.

Independence from the statewide toggle:
- With the cutoff still set to 4, set `CardReplacement.Enabled` to `false` and restart. No replacement CTA or per-card links for anyone, including Sofia or Luna. Address update still works.
- Turn `Enabled` back to `true`. Cutoff behavior returns.

Regression check: 
- Comment the setting back out (or delete it) and restart. All three children can request replacement again, subject to the usual issuance-type and card-status rules.

[DC-742]: https://codeforamerica.atlassian.net/browse/DC-742?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


spokenbird (approved):
Looks good! Tested locally and confirmed works as intended.

PR #678, opened 2026-09-11
[DC-691] Fail startup when OIDC is enabled with an incomplete client

adbergen:
#### 🔗 Jira ticket

[DC-691 — Add typed OidcSettings with validate-on-start for base OIDC config](https://codeforamerica.atlassian.net/browse/DC-691)

#### ✍️ Description

**Most of this ticket already shipped.** `OidcSettings` exists in `Core.AppSettings`, binds the `Oidc` section, and is registered via `AddPortalOptions<OidcSettings>()` — which already wraps `AddOptionsWithValidateOnStart` plus `ValidateDataAnnotations`. Validators auto-register by reflection over `IValidateOptions<T>`, so `OidcSettingsValidator` was already wired in. Every controller, service and middleware the ticket names had already moved off raw string keys; their remaining `Oidc:` mentions are comments and error-message text.

What was missing is the part the ticket is actually about.

**The validator only enforced signing-key length.** It never required the rest of the client. A deployment with `Oidc:DiscoveryEndpoint` set but no `ClientId` booted cleanly and failed only once a visitor came back from the IdP — exactly the "login looks fine, then breaks after redirect" symptom in the ticket.

Now, when a discovery endpoint is present, `ClientId`, `CallbackRedirectUri` and a 32-character `CompleteLoginSigningKey` are all required and a gap fails the boot. An absent discovery endpoint still means OIDC is off and nothing else is required, matching how the state allowlist already reads that key. Failures collect every problem into one message so a single boot tells an operator everything to fix, following the existing `OidcStepUpSettingsValidator` pattern.

**Worth a close look during review: an existing test was asserting the bug.** `Startup_WithEmptyOidcSigningKey_DoesNotThrow` claimed, in its remark, to cover states without OIDC. But the shared fixture sets a real `Oidc__DiscoveryEndpoint`, so it was really asserting that an OIDC-enabled instance with no signing key starts happily. It has been split into the behaviour it claimed to cover (clear discovery, empty section, still boots) and the behaviour it actually exercised (OIDC on, missing key, now throws), plus startup cases for the other two required fields.

**Also migrated three startup-time raw `config["Oidc:…"]` reads** to bind `OidcSettings` off the section, so property names are compile-checked. These run before the DI container exists, so `IOptions<OidcSettings>` genuinely isn't resolvable at those call sites; binding straight from `IConfiguration` also keeps the step-up validator off the options system, so validating one options type can't pull in another.

**Deliberately not done.** Nesting is unchanged — `OidcStepUpSettings` and `OidcVerificationClaimSettings` stay as sibling options types, which the ticket allowed. Because of that the two `Oidc:StepUp:*` reads in `IdProofingRequirementsCoherenceValidator` are left alone; the ticket made touching those conditional on nesting changing. `LanguageParam` is not added: it appears in no source appsettings file and is read by no code, so the only trace of it is a stale `bin/Release` artifact.

#### ⚠️ Deploy note for the reviewer

This can turn a currently-starting app into a hard boot failure, which is the intent — but it is worth confirming against real config before merge.

Real state configs are gitignored, so only the examples are visible here. `appsettings.co.example.json` has every field populated and `appsettings.dc.example.json` has no `Oidc` block at all, so both pass. If a live environment has `DiscoveryEndpoint` set while missing `ClientId`, `CallbackRedirectUri` or the signing key, it will now fail to start rather than fail at login. That environment's login flow is already broken today; this only moves the failure somewhere an operator can see it.

#### 🔗 Links to related PRs

None. Branch is cut from `main` and touches only OIDC configuration validation.

#### ✅ Completion tasks

- [x] Added relevant tests — 19 validator unit tests (up from 7) and 5 startup-validation integration tests (up from 2), covering each required field individually, the every-problem-at-once message, and the OIDC-off path.
- [ ] Meets acceptance criteria in ticket — mostly, with two deliberate gaps explained above: nesting is unchanged, and `LanguageParam` is not added because nothing reads it. Leaving this for the reviewer to judge.
- [x] Configuration changes: N/A — no new env vars, secrets, or appsettings. **But see the deploy note above:** existing `Oidc:*` values are now required when OIDC is enabled, so a live environment missing one will fail to start. Someone with access to the real DC and CO config should confirm before merge.


noxmwalsh (approved):
Sorry for the delay on this one, but I think this is good to go!

PR #680, opened 2026-09-14
[SEBT-700] Ask whether users receive SNAP or TANF before choosing an ID, behind a feature flag

adbergen:
#### 🔗 Jira ticket

[SEBT-700 — As a DC co-loaded user I want to be able to select for this so I have a clear pathway to logging into the SEBT Portal](https://codeforamerica.atlassian.net/browse/SEBT-700)

#### ✍️ Description

Behind a new feature flag, **`enable_socure_snap_tanf_question`** (off by default), the Socure id-proofing form asks **"Do you receive SNAP or TANF?"** right after date of birth, matching the design screenshots from the ticket.

- **Yes** shows only the SNAP or TANF case number field (7 or 8 digits).
- **No** shows "Which ID number can you provide?" with SSN, ITIN, and None of the above.

With the flag **off**, the form is exactly what ships today: the four ID options, the narrower list for returning co-loaded users, and a required ID choice.

**Why a flag, and why this shape.** Which state turns this on is still open, so the flag is named for the Socure id-proofing path rather than a state, and it lives only in the base `appsettings.json`. A state opts in through its own appsettings or AWS AppConfig. The change is kept additive so that reverting it, or deleting the flag after rollout, touches little:

- The existing option lists, their tests, and every e2e spec are unchanged from `main`.
- The flag is read in one place, `IdProofingWithDi`, which picks the question design or the original options.
- The form learns an optional `snapTanfOption` prop. Without it, the form renders and validates exactly as before.
- Net diff is +456 / −30 across 7 files.

**With the flag on, only date of birth is required**, confirmed with design. Leaving the question unanswered, or answering Yes with a blank case number, submits with no ID, which the backend already handles the same way as "None of the above". Anything entered is still shape-checked, and a chosen SSN or ITIN still needs its number, since "None of the above" is the explicit way out on that path.

**The backend is unchanged.** Yes submits `snapAccountId`; No submits `ssn`, `itin`, or no ID, exactly as the form already could.

**With the flag on, co-loaded users get the question too.** The original options come from `session.isCoLoaded`, but that flag only turns true after a SNAP/TANF match, so a first-time co-loaded user, the group this ticket is about, arrives with it false. Everyone answers the same question instead.

**The form holds while flags load**, so nobody starts filling in one design and has it swapped for the other. Flags are normally cached by the time a user reaches id-proofing after sign-in.

**Analytics:** with the flag on, a failed "Yes" submission that includes a case number is tagged `idv_primary_reason: not_found` (a SNAP/TANF lookup miss, no Socure) rather than `socure_fail`. A "Yes" with the case number left blank submits no ID, so a Socure failure there still reports `socure_fail`.

#### ⚠️ Content gap

The helper text under the case number field in the design ("This number can be found on your SNAP or TANF notices. It can also be found on the District Direct website or mobile app. It is typically 7 or 8 digits long.") is not in the DC content sheet yet. The field reads the `helperAccountId` key and renders the hint only once that key exists, so it shows no hint until a row lands (for example `S8 - ID Proofing Optional ID Info - Helper Account ID`), then picks it up with no code change. Provisional English was not an option: its guard test restricts it to dashboard status labels.

#### 🔗 Links to related PRs

None. Branch is cut from `main` and touches only the id-proofing form, its option lists, and one feature flag default.

#### ✅ Completion tasks

- [x] Added relevant tests: form tests cover the question, both paths, optional submission, case number validation, not carrying a value across answers, the content-gated helper, and failure analytics; the original form tests still run against the flag-off form unchanged. `IdProofingWithDi` tests keep the original option-set cases for flag off and add flag-on cases for every session state plus the loading hold. Backend feature flag tests pass with the new default. E2e specs are unchanged and exercise the flag-off form.
- [ ] Meets acceptance criteria in ticket: met with the flag on; off by default until a state is chosen.
- [ ] Configuration changes: one new feature flag, `enable_socure_snap_tanf_question`, defaulting to `false` in base `appsettings.json`. Not added to any state example file or AWS AppConfig until a state is chosen.


apps/portal/src/SEBT.Portal.Web/src/features/auth/components/id-proofing/IdProofingForm.tsx
```diff
@@ -261,11 +291,12 @@ export function IdProofingForm({ idOptions, contactLink, getDiToken }: IdProofin
...
         } else {
-          // Co-loaded users reach "failed" only via SNAP/TANF + DOB mismatch (no Socure),
-          // or when the backend classified the household as co-loaded-only.
+          // Co-loaded users, and anyone who answered "Yes" to SNAP/TANF, reach "failed" only via
+          // SNAP/TANF + DOB mismatch (no Socure), or when the backend classified the household as
+          // co-loaded-only.
           setPageData(
             'idv_primary_reason',
-            isCoLoaded || response.offboardingReason === 'coLoadedOnly'
+            isCoLoaded || answeredYes || response.offboardingReason === 'coLoadedOnly'
```
necampanini: i think there's an edge cases where a user who answers "Yes" but leaves the case number blank still goes to Socure. if they return a failure there, this would still get tagged `not_found`.

any way to account for this so a blank "Yes" that Socure fails on would return `socure_fail`?

necampanini (changes requested):
looks good, just the one edge case we may need to consider

apps/portal/src/SEBT.Portal.Web/src/features/auth/components/id-proofing/IdProofingForm.tsx
adbergen: Addressed in [15c0fae](https://github.com/codeforamerica/sebt-self-service-portal/pull/680/commits/15c0faefd40a061fd78a4201bd5e4cd1770497d3)

necampanini (approved):

PR #696, opened 2026-09-17
[SEBT-692] Validate remaining portal settings at startup

adbergen:
#### 🔗 Jira ticket

[SEBT-692 — Align options registration so remaining AppSettings validate on start](https://codeforamerica.atlassian.net/browse/SEBT-692)

#### ✍️ Description

**Registration was already aligned.** The ticket describes settings bound with plain `AddOptions` in `AddPortalInfrastructureAppSettings`. That method no longer exists: #587 moved every settings type onto `AddPortalOptions<T>()` in `Infrastructure/Configuration/Dependencies.cs`, which registers `AddOptionsWithValidateOnStart`, runs `ValidateDataAnnotations`, and auto-registers any `IValidateOptions<T>` in the assembly. Every type the ticket lists already validates on start, and `IOptionsMonitor` consumers re-validate on AppConfig reload. What was missing was rules for those types to enforce, so that is what this adds.

**Decision for each type the ticket lists:**

| Settings | Decision | Why |
|---|---|---|
| `RedisSettings` | **New `RedisSettingsValidator`** | Once `Host` is set: port must be 1–65535, `SslHost` and `AcceptSelfSignedCertificates` require `Ssl` (without TLS they are ignored and the connection is unencrypted), and self-signed certificates fail outside Development. A blank `Host` means Redis is off and nothing else is checked: blanking the host over a state file that configures TLS Redis is how the CO integration e2e stack switches to in-memory stores. |
| `StateHouseholdIdSettings` | `[MinLength(1)]` | With no preferred type, no guardian links to a household. |
| `IdProofingValiditySettings` | `[Range(1, …)]` on `ValidityDays` | Zero or negative expires every completed verification immediately. |
| `AddressValidationDataSettings` | `[Range(0, …)]` on `MaxStreetAddressLength` | 0 still disables the check; negative is nonsense. |
| `OidcVerificationClaimSettings` | `[Required]` on the two primary claim names | A blank claim name matches no claim, so every OIDC user reads as unverified. Blank fallbacks keep meaning "use the default". |
| `EnrollmentCheckerSettings` | **New `EnrollmentCheckerSettingsValidator`**, income figures only | Negative figures fail, and `BaseThreshold` and `MaxHouseholdSize` must be set together (with one, the checker silently drops income screening). **Banner copy is deliberately unvalidated:** this section hot-reloads, and a rejected reload fails every consumer until corrected, including the checker's whole features endpoint (outage page, apply, season). A blank banner does not justify that. A test pins this choice. |
| `SeedingSettings` | Left unvalidated | Dev-only, and a bad `EmailPattern` already fails loudly at seed time. |
| `IdProofingEligibilitySettings`, `CoLoadedCohortFilterSettings`, `AddressValidationPolicySettings` | Left unvalidated | Booleans only; the binder already rejects non-boolean values. |
| `FeatureManagementSettings` | N/A | No such type. Flags go through Microsoft.FeatureManagement. |

**Redis is also checked where it is read early.** `ResolveRedisConfigurationOptions` reads the section while services are being registered, before validate-on-start runs, to wire the distributed cache and lock provider. Its inline self-signed-certificate check is replaced by a call to the same validator, so that path now applies every Redis rule and throws the same collected message as an `OptionsValidationException` instead of an `InvalidOperationException`. The one existing test that expected the old type is updated.

Validators follow the existing `OidcSettingsValidator` and `PiiEncryptionProductionSecretsValidator` patterns: problems are collected so one boot reports everything to fix, and environment-dependent rules take `IHostEnvironment`.

#### ⚠️ Deploy note for the reviewer

These rules turn some currently-booting configurations into startup failures, which is the intent, but real state configs are gitignored and could not be checked here. Please confirm no live DC or CO environment has:

- a `Redis:Host` set together with `SslHost` or `AcceptSelfSignedCertificates` while `Ssl` is off
- `Redis:AcceptSelfSignedCertificates: true` outside Development (this already failed before this PR)
- `EnrollmentChecker:IncomeEligibility` with only one of `BaseThreshold` / `MaxHouseholdSize` set, or a negative figure (including in the AWS AppConfig AppSettings profile, where a bad value would now be rejected on reload)
- an empty `StateHouseholdId:PreferredHouseholdIdTypes`

All example appsettings files in the repo pass.

#### 🔗 Links to related PRs

- #587 aligned the registrations this builds on.
- #678 added the collected-problems `OidcSettingsValidator` pattern used here.

#### ✅ Completion tasks

- [x] Added relevant tests: `RedisSettingsValidatorTests` and `EnrollmentCheckerSettingsValidatorTests` cover success and failure for each rule, including defaults and the example-file shapes; `SettingsDataAnnotationsTests` covers each new data annotation through the same `DataAnnotationValidateOptions` that `ValidateDataAnnotations` registers; `RedisSettingsStartupValidationTests` proves a production host with self-signed Redis certificates refuses to start. Full backend unit suite passes.
- [x] Meets acceptance criteria in ticket: documented decision per type above, high-impact settings fail startup or reload, empty defaults still boot, registration already consistent via #587.
- [ ] Configuration changes: N/A, no new settings, secrets, or env vars. **But see the deploy note:** existing values are now checked, so a live environment with one of the listed shapes will fail to start.


apps/portal/src/SEBT.Portal.Infrastructure/Configuration/Validators/RedisSettingsValidator.cs
```diff
@@ -0,0 +1,63 @@
...
+/// Redis is optional: without <c>Redis:Host</c> the portal falls back to in-memory caching and SQL
+/// Server locks. Blanking the host over a state file that configures TLS Redis is how a
+/// single-instance stack turns Redis off, so a section without a host is not checked. Once a host is
+/// set, the TLS values have to agree with each other, and certificate validation may only be skipped
+/// in Development.
+///
+/// <c>ResolveRedisConfigurationOptions</c> runs this too: it reads the section while services are still
+/// being registered, before validate-on-start can see it.
+/// </summary>
+public class RedisSettingsValidator(IHostEnvironment environment) : IValidateOptions<RedisSettings>
```
noxmwalsh, line 19: This is minor, but just to match `OidcSettingsValidator`, can we add some `null` options type guards within this validator?

noxmwalsh (approved):
Approved with comment about adding some null guards for the Redis settings validator

PR #697, opened 2026-09-17
[SEBT-761] Add one-command local environment setup script

adbergen:
#### 🔗 Jira ticket

[SEBT-761 — Create one-click script to install / set up whole local env](https://codeforamerica.atlassian.net/browse/SEBT-761)

#### ✍️ Description

Adds one command that takes a fresh clone to a machine ready to run the portal, automating the README's manual steps 1 through 5:

```bash
./scripts/dev/setup-local.sh              # both states
./scripts/dev/setup-local.sh --state co   # one state
pnpm dev:setup                            # same thing, once pnpm exists
```

**What it does, in order.** Every step checks before it acts, so re-running is safe.

1. **Platform.** macOS with Homebrew; anything else points to the manual steps in the README.
2. **Tools.** Installs whatever is missing: .NET 10 (`brew install dotnet`, then confirms a 10.x SDK for `global.json`), Node.js 24 or newer (`brew install node`), and pnpm 10 (`npm install -g pnpm@10`, since the team stays on 10 and brew's latest is 11). Each is re-checked after installing, so a different version first on PATH stops setup with a clear message instead of failing later. **Docker Desktop is checked, not installed**: installing it needs admin rights and it has to be opened once from the GUI, so setup stops with the exact steps if it is missing or not running.
3. **DC connector.** Clones `sebt-self-service-portal-dc-connector` beside this repo if it is not there. The clone runs over HTTPS with credential prompts off, so without access it fails fast and setup continues for CO only with a warning. An existing checkout is never switched or pulled, but setup warns when it is off main or behind, since an old branch can fail to build against current shared package versions.
4. **Config.** `scripts/dev/setup-local-config.sh` creates `.env`, the web `.env.local`, and the Development and per-state appsettings from their examples, **never overwriting an existing file**. In files it creates, it fills the database password from `.env` (so compose and the API agree) and generates the JWT, identifier hasher, and OIDC signing keys. Values only a developer can supply, such as CBMS and OIDC client IDs, stay as placeholders and are listed at the end.
5. **Dependencies.** `pnpm install` and `dotnet tool restore`.
6. **Build.** The existing `build-dc.sh` (which also builds the monorepo) or `build-co.sh`.
7. **Services.** Generates the Redis certs and runs `docker compose up -d --wait` for MSSQL, Mailpit and Redis, waiting until they report healthy.
8. **Summary.** The command to start the app per state, the app and Mailpit URLs, a pointer to the local OIDC docs for CO, and the placeholders still to fill in. It stops there rather than launching the dev server, so setup stays re-runnable and separate from the long-running process.

The README's Local Environment Set Up section gets a short "Quick start (macOS)" pointing at the script; the manual steps stay below it.

**Verified on a real machine:**
- `--state co` end to end: tools detected, config files created with placeholders listed, CO build succeeded, MSSQL, Mailpit and Redis came up healthy, summary printed.
- Re-run: every config file kept, generated JWT key unchanged, build and services still fine.
- `--help`, an unknown `--state`, shellcheck, and parsing under macOS's system bash 3.2.
- The DC path on the test machine stopped at the build because that machine's DC connector checkout was on an old feature branch, 13 commits behind its main, which is exactly the case the new warning calls out; `pnpm dev:dc` fails the same way there.

#### 🔗 Links to related PRs

None. Branch is cut from `main` and touches only developer scripts, one `package.json` alias, and the README.

#### ✅ Completion tasks

- [x] Added relevant tests: `scripts/ci/tests/setup-local-config_test.sh` runs the config step against a temporary copy of the real example files, so it breaks if their placeholders change. It covers a fresh checkout, distinct 32+ character generated keys, a re-run keeping existing files, `--state co` skipping the DC file, and rejecting an unknown state. It runs in the existing CI scripts smoke tests job. The install, clone, build and Docker steps change the machine, so they were verified by running the script rather than in CI.
- [ ] Configuration changes: N/A


zarahzachz (changes requested):
This work is _fantastic_! 🤩 

I did run into problems running the script locally where the script bailed out before running the docker commands. 

I think this due to a pnpm versioning issue. I troubleshat it and found adding a line like `"packageManager": "pnpm@10.0.0",` to the root package.json file and running `pnpm install` before rerunning this shellscript seemed to shake out the issues and allowed the whole script to run to completion.

